PR #22, opened 2015-06-04
Tokenizer inside Liquid::C

gauravmc:
@dylanahsmith Is this what you had in mind in your comment [here](https://github.com/Shopify/liquid/pull/590#discussion_r31670239) regarding defining Tokenizer such that it doesn't conflict with ruby `Liquid::Tokenizer` until it supports line numbers? cc @pushrax


dylanahsmith:
:+1:

trishume:
:+1:

PR #24, opened 2015-07-05
Implement line numbers support.

dylanahsmith:
@pushrax for review

Depends on https://github.com/Shopify/liquid/pull/614

Implement line number calculation in the liquid extension so that we don't need to fallback to ruby liquid when using the `line_numbers: true` option.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -11,7 +11,8 @@ static ID
     intern_clear,
     intern_tags,
     intern_parse,
-    intern_square_brackets;
+    intern_square_brackets,
+    intern_set_line_number;
 
 static int is_id(int c)
 {
@@ -34,6 +35,9 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
     VALUE nodelist = rb_ivar_get(self, intern_nodelist);
 
     while (true) {
+        if (tokenizer->line_number != 0) {
+            rb_funcall(options, intern_set_line_number, 1, UINT2NUM(tokenizer->line_number));
+        }
         tokenizer_next(tokenizer, &token);
 
         switch (token.type) {
@@ -115,6 +119,7 @@ void init_liquid_block()
     intern_tags = rb_intern("tags");
     intern_parse = rb_intern("parse");
     intern_square_brackets = rb_intern("[]");
+    intern_set_line_number = rb_intern("line_number=");
 
     VALUE cLiquidBlockBody = rb_const_get(mLiquid, rb_intern("BlockBody"));
     rb_define_method(cLiquidBlockBody, "c_parse", rb_block_parse, 2);
--- a/ext/liquid_c/tokenizer.c
+++ b/ext/liquid_c/tokenizer.c
@@ -38,7 +38,7 @@ static VALUE tokenizer_allocate(VALUE klass)
     return obj;
 }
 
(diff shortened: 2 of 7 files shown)


dylanahsmith:
Benchmark with line number supports on master

```
              parse:       22.8 (±4.4%) i/s -       1366 in  60.052357s
        parse & run:       18.9 (±5.3%) i/s -       1129 in  60.036080s
```

on this branch

```
              parse:       65.7 (±4.6%) i/s -       3930 in  59.999107s
        parse & run:       23.3 (±4.3%) i/s -       1398 in  60.059979s
```

performance without line number support was comparable between master and this branch.  Here is the benchmark results on master:

```
              parse:       69.3 (±7.2%) i/s -       4140 in  60.069889s
        parse & run:       23.3 (±4.3%) i/s -       1390 in  59.997252s
```

and on this branch

```
              parse:       69.4 (±5.8%) i/s -       4152 in  60.038627s
        parse & run:       23.2 (±4.3%) i/s -       1390 in  60.002630s
```

ext/liquid_c/block.c
```diff
@@ -34,6 +35,9 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
     VALUE nodelist = rb_ivar_get(self, intern_nodelist);
 
     while (true) {
+        if (tokenizer->line_number != 0) {
```
pushrax: Why is this conditional needed?

pushrax: Oh, it's a boolean. Let's use true and false instead?

pushrax: Oh, it's not a boolean, forget everything I said. This code is a bit confusing but makes sense, how about adding a comment somewhere saying that it uses `line_number` as a flag as well as a number?

ext/liquid_c/tokenizer.c
```diff
@@ -48,6 +48,7 @@ static VALUE tokenizer_initialize_method(VALUE self, VALUE source)
     tokenizer->source = source;
     tokenizer->cursor = RSTRING_PTR(source);
     tokenizer->length = RSTRING_LEN(source);
+    tokenizer->line_number = RTEST(line_numbers) ? 1 : 0;
```
pushrax: Perhaps this would be a good place for the comment I mentioned.

ext/liquid_c/tokenizer.c
```diff
@@ -114,6 +115,16 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
+
+    if (tokenizer->line_number) {
+        const char *cursor = token->str;
+        const char *end = token->str + token->length;
+        while (cursor < end) {
+            if (*cursor == '\n')
+                tokenizer->line_number++;
+            cursor++;
+        }
+    }
```
pushrax: I would prefer if we could do this while iterating the first time, but I also realize the code would be a little messy and prone to missing cases :(

pushrax:
:+1: - thank you for taking care of this, it's great to kill that inconsistency with Ruby Liquid.

ext/liquid_c/tokenizer.c
dylanahsmith: It would involve doing a check at a number of places in this method (https://github.com/shopify/liquid-c/commit/4151be34ea3ebd46dc07d6e82152b3b9f37d299b) and would also add unnecessary overhead when line numbers aren't getting used.

PR #28, opened 2015-07-31
Don't allow identifiers to end in a dash in strict parse mode.

dylanahsmith:
@pushrax & @trishume for review

This just ports https://github.com/Shopify/liquid/pull/636 to liquid-c

### DIFF
--- a/ext/liquid_c/lexer.c
+++ b/ext/liquid_c/lexer.c
@@ -134,6 +134,7 @@ const char *lex_one(const char *start, const char *end, lexer_token_t *token)
     if (ISALPHA(c) || c == '_') {
         cur = str;
         while (++cur < end && is_identifier(*cur)) {}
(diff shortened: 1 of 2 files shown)


pushrax:
:+1: pending https://github.com/Shopify/liquid/pull/636

ext/liquid_c/lexer.c
```diff
@@ -134,6 +134,7 @@ const char *lex_one(const char *start, const char *end, lexer_token_t *token)
     if (ISALPHA(c) || c == '_') {
         cur = str;
         while (++cur < end && is_identifier(*cur)) {}
+        if (*(cur - 1) == '-') cur--;
```
pushrax: Pretty sure this will only strip off the last `-`.

dylanahsmith: good point, this should be a while loop

PR #30, opened 2016-06-27
Add support for whitespace control character.

evulse:
This matches up with Shopify/liquid#746 to resolve issues Shopify/liquid#216, Shopify/liquid#215, Shopify/liquid#214, Shopify/liquid#194, Shopify/liquid#171, Shopify/liquid#162

This pull request is designed to be standalone and will pass all tests without its counterpart Shopify/liquid#746 however Shopify/liquid#746 requires this pull request and will be updated with this dependancy once merged.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -63,9 +63,7 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_RAW:
             {
-                const char *start = token.str, *end = token.str + token.length;
-                const char *token_start = start;
-                const char *token_end = end;
+                const char *start = token.str, *end = token.str + token.length, *token_start = start, *token_end = end;
 
                 if(token.lstrip)
                     token_start = read_while(start, end, rb_isspace);
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -56,6 +56,10 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             case TOKEN_RAW:
             {
                 VALUE str = rb_enc_str_new(token.str, token.length, utf8_encoding);
+
+                if(token.trim_whitespace)
+                   rb_funcall(str, rb_intern("lstrip!"), 0);
+
                 rb_ary_push(nodelist, str);
 
                 if (rb_ivar_get(self, intern_blank) == Qtrue) {
@@ -68,7 +72,22 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str + 2, token.length - 4, utf8_encoding), options};
+                const char *start = token.str + 2;
+                long end = token.length - 4;
+
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
+                    rb_ary_push(nodelist, last_node);
+                    start++;
+                    end--;
+                }
+
+                if (token.str[token.length - 3] == '-') {
+                    end--;
+                }
+
+                VALUE args[2] = {rb_enc_str_new(start, end, utf8_encoding), options};
                 VALUE var = rb_class_new_instance(2, args, cLiquidVariable);
                 rb_ary_push(nodelist, var);
                 rb_ivar_set(self, intern_blank, Qfalse);
@@ -77,6 +96,17 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             case TOKEN_TAG:
             {
                 const char *start = token.str + 2, *end = token.str + token.length - 2;
+
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
+                    rb_ary_push(nodelist, last_node);
+                    start++;
+                }
+
+                if (token.str[token.length - 3] == '-') {
+                     end--;
+                }
 
                 // Imitate \s*(\w+)\s*(.*)? regex
                 const char *name_start = read_while(start, end, rb_isspace);
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -59,10 +59,12 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
                 const char *start = token.str, *end = token.str + token.length;
                 const char *token_start = start;
 
-                if(token.trim_whitespace)
+                if(token.lstrip)
                     token_start = read_while(start, end, rb_isspace);
 
                 VALUE str = rb_enc_str_new(token_start, end - token_start, utf8_encoding);
+                if(token.rstrip)
+                    rb_funcall(str, intern_rstrip, 0);
                 rb_ary_push(nodelist, str);
 
                 if (rb_ivar_get(self, intern_blank) == Qtrue) {
@@ -75,23 +77,7 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_VARIABLE:
             {
-                const char *start = token.str + 2;
-                long length = token.length - 4;
-
-                if (token.str[2] == '-') {
-                    if (RARRAY_LEN(nodelist) > 0) {
-                        VALUE last_node = RARRAY_AREF(nodelist, RARRAY_LEN(nodelist)-1);
-                        rb_funcall(last_node, intern_rstrip, 0);
-                    }
-                    start++;
-                    length--;
-                }
-
-                if (token.str[token.length - 3] == '-') {
-                    length--;
-                }
-
-                VALUE args[2] = {rb_enc_str_new(start, length, utf8_encoding), options};
+                VALUE args[2] = {rb_enc_str_new(token.str + 2 + token.lstrip, token.length - 4 - token.lstrip - token.rstrip, utf8_encoding), options};
                 VALUE var = rb_class_new_instance(2, args, cLiquidVariable);
                 rb_ary_push(nodelist, var);
                 rb_ivar_set(self, intern_blank, Qfalse);
@@ -99,19 +85,7 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_TAG:
             {
-                const char *start = token.str + 2, *end = token.str + token.length - 2;
-
-                if (token.str[2] == '-') {
-                    if (RARRAY_LEN(nodelist) > 0) {
-                        VALUE last_node = RARRAY_AREF(nodelist, RARRAY_LEN(nodelist)-1);
-                        rb_funcall(last_node, intern_rstrip, 0);
-                    }
-                    start++;
-                }
-
-                if (token.str[token.length - 3] == '-') {
-                     end--;
-                }
+                const char *start = token.str + (2 + token.lstrip), *end = token.str + (token.length - 2 - token.rstrip);
 
                 // Imitate \s*(\w+)\s*(.*)? regex
                 const char *name_start = read_while(start, end, rb_isspace);
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -12,8 +12,7 @@ static ID
     intern_registered_tags,
     intern_parse,
     intern_square_brackets,
-    intern_set_line_number,
-    intern_rstrip;
+    intern_set_line_number;
 
 static int is_id(int c)
 {
@@ -24,6 +23,14 @@ inline static const char *read_while(const char *start, const char *end, int (fu
 {
     while (start < end && func((unsigned char) *start)) start++;
     return start;
+}
+
+inline static const char *read_while_end(const char *start, const char *end, int (func)(int))
+{
+    end--;
+    while (start < end && func((unsigned char) *end)) end--;
+    end++;
+    return end;
 }
 
 static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
@@ -58,13 +65,14 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             {
                 const char *start = token.str, *end = token.str + token.length;
                 const char *token_start = start;
+                const char *token_end = end;
 
                 if(token.lstrip)
                     token_start = read_while(start, end, rb_isspace);
-
-                VALUE str = rb_enc_str_new(token_start, end - token_start, utf8_encoding);
                 if(token.rstrip)
-                    rb_funcall(str, intern_rstrip, 0);
+                    token_end = read_while_end(token_start, end, rb_isspace);
+                    
+                VALUE str = rb_enc_str_new(token_start, token_end - token_start, utf8_encoding);
                 rb_ary_push(nodelist, str);
 
                 if (rb_ivar_get(self, intern_blank) == Qtrue) {
@@ -129,7 +137,6 @@ void init_liquid_block()
     intern_parse = rb_intern("parse");
     intern_square_brackets = rb_intern("[]");
     intern_set_line_number = rb_intern("line_number=");
-    intern_rstrip = rb_intern("rstrip!");
 
     VALUE cLiquidBlockBody = rb_const_get(mLiquid, rb_intern("BlockBody"));
     rb_define_method(cLiquidBlockBody, "c_parse", rb_block_parse, 2);
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -91,7 +91,7 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_TAG:
             {
-                const char *start = token.str + (2 + token.lstrip), *end = token.str + (token.length - 2 - token.rstrip);
+                const char *start = token.str + 2 + token.lstrip, *end = token.str + token.length - 2 - token.rstrip;
 
                 // Imitate \s*(\w+)\s*(.*)? regex
                 const char *name_start = read_while(start, end, rb_isspace);
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -12,7 +12,8 @@ static ID
     intern_registered_tags,
     intern_parse,
     intern_square_brackets,
-    intern_set_line_number;
+    intern_set_line_number,
+    intern_rstrip;
 
 static int is_id(int c)
 {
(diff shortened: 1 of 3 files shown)


fw42:
ping @Shopify/liquid

evulse:
This adds support for {{- and {%- syntax which will lstrip! and -}} and -%} which will rstrip!

evulse:
Now matches up to Shopify/liquid#773

tobi:
I'd love to avoid the cycle through the Ruby VM to call rstrip!/lstrip!. 

Ideally we should parse white-space as a special token, then modify the token to be marked for render-skipping if {%- or -%} are encountered.

tobi:
I see now that this would make it harder to match with Shopify/liquid#746 - alright, let's not overthink it. Thanks!

evulse:
Yeah I would have liked to avoid it but the trade off isn't worth it at this stage. To keep the functions inline on both sides makes maintenance easier. Also as I'm pulling back the last token if needed it requires conversion from a ruby object then back again where as this approach keeps it clean and only a few lines

ext/liquid_c/block.c
```diff
@@ -56,6 +56,10 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             case TOKEN_RAW:
             {
                 VALUE str = rb_enc_str_new(token.str, token.length, utf8_encoding);
+
+                if(token.trim_whitespace)
+                   rb_funcall(str, rb_intern("lstrip!"), 0);
```
pushrax: Why not find the offset after any whitespace (see [lstrip_offset](https://github.com/ruby/ruby/blob/c9df0f82b7f364e95a9cf8a9492b00759cfe61af/string.c#L7926)) and create the string with that offset?

ext/liquid_c/block.c
```diff
@@ -68,7 +72,22 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
...
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str + 2, token.length - 4, utf8_encoding), options};
+                const char *start = token.str + 2;
+                long end = token.length - 4;
+
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
+                    rb_ary_push(nodelist, last_node);
```
pushrax: Just use `rb_ary_last` instead of popping and pushing.

ext/liquid_c/block.c
```diff
@@ -68,7 +72,22 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str + 2, token.length - 4, utf8_encoding), options};
+                const char *start = token.str + 2;
+                long end = token.length - 4;
+
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
```
pushrax: I don't mind `rstrip!` because it just sets the length – no memmove required.

ext/liquid_c/block.c
```diff
@@ -78,6 +97,17 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             {
                 const char *start = token.str + 2, *end = token.str + token.length - 2;
 
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
+                    rb_ary_push(nodelist, last_node);
```
pushrax: ditto for `rb_ary_last`

ext/liquid_c/block.c
```diff
@@ -68,7 +72,22 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str + 2, token.length - 4, utf8_encoding), options};
+                const char *start = token.str + 2;
+                long end = token.length - 4;
```
pushrax: Call this `length` instead.

ext/liquid_c/block.c
pushrax: I'm now thinking that it could be a good idea to make `read_while` respect the codepoint length (see other usages of `read_while` for what it does).

ext/liquid_c/tokenizer.c
```diff
@@ -127,6 +129,13 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
             cursor++;
         }
     }
+
+    if(token->type == TOKEN_RAW && tokenizer->trim_whitespace ) {
```
pushrax: spacing

ext/liquid_c/block.c
```diff
@@ -78,6 +97,17 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             {
                 const char *start = token.str + 2, *end = token.str + token.length - 2;
 
+                if (token.str[2] == '-') {
+                    VALUE last_node = rb_ary_pop(nodelist);
+                    rb_funcall(last_node, rb_intern("rstrip!"), 0);
```
pushrax: We've been typically caching the result of `rb_intern`

ext/liquid_c/block.c
```diff
@@ -56,6 +56,10 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             case TOKEN_RAW:
             {
                 VALUE str = rb_enc_str_new(token.str, token.length, utf8_encoding);
+
+                if(token.trim_whitespace)
```
pushrax: spacing

pushrax:
Makes sense and looks pretty good overall.

ext/liquid_c/block.c
evulse: Agreed this is a much better approach, implementing now

evulse: It looks like rb_ary_last isn't exposed so have used some of its logic directly

ext/liquid_c/block.c
```diff
@@ -68,7 +75,23 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
             }
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str + 2, token.length - 4, utf8_encoding), options};
+                const char *start = token.str + 2;
+                long length = token.length - 4;
+
+                if (token.str[2] == '-') {
```
dylanahsmith: You can use `start[0]` or `*start` instead

ext/liquid_c/block.c
```diff
@@ -68,7 +75,23 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
...
+                    if (RARRAY_LEN(nodelist) > 0) {
+                        VALUE last_node = RARRAY_AREF(nodelist, RARRAY_LEN(nodelist)-1);
+                        rb_funcall(last_node, intern_rstrip, 0);
+                    }
+                    start++;
+                    length--;
+                }
+
+                if (token.str[token.length - 3] == '-') {
```
dylanahsmith: I would use `if (length > 0 && start[length - 1] == '-') {` because I am concerned about a case like `{{-}}` would cause length to go negative, which is a scary state to be in C code.

ext/liquid_c/block.c
```diff
@@ -78,6 +101,18 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
...
 
+                if (token.str[2] == '-') {
+                    if (RARRAY_LEN(nodelist) > 0) {
+                        VALUE last_node = RARRAY_AREF(nodelist, RARRAY_LEN(nodelist)-1);
+                        rb_funcall(last_node, intern_rstrip, 0);
+                    }
+                    start++;
+                }
+
+                if (token.str[token.length - 3] == '-') {
```
dylanahsmith: Same concern as above, where `{%-%}` can cause the end pointer to come before the start pointer.

ext/liquid_c/tokenizer.c
```diff
@@ -127,6 +128,10 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
             cursor++;
         }
     }
+    if(token->type == TOKEN_RAW && tokenizer->trim_whitespace) {
+        token->trim_whitespace = 1;
+    }
+    tokenizer->trim_whitespace = ((token->type == TOKEN_VARIABLE || token->type == TOKEN_TAG) && token->str[token->length - 3] == '-');
```
dylanahsmith: I don't see why you are doing these checks down here when you could do them in context where token->type is set.

That would also allow you to easily detect if the next token should be trimmed, since the function finds the end of a raw string by finding the start of the next tag or variable.  That means you could avoid having to construct a ruby string that needs to be right stripped.  You could store that as two flags on the token for whether the raw tag should be left and/or right trimmed.

ext/liquid_c/tokenizer.h
```diff
@@ -13,13 +13,15 @@ typedef struct token {
     enum token_type type;
     const char *str;
     long length;
+    unsigned int trim_whitespace;
```
dylanahsmith: If you move this field after `enum token_type type;` then it would avoid unnecessary padding on x64, since right now it stores 4 bytes for the enum followed by 4 bytes of padding, then two 8 byte fields for `str` and `length`.

ext/liquid_c/block.c
dylanahsmith: I also think we should avoid this code duplication using an inline function so we can solve this problem once.  This function could also be defined with the tokenizer code so whitespace trimming can be handled in one place.  There could be one function to get the tag token contents (i.e. without the tag start, end or trim characters) and another to get the raw token contents (i.e. after trimming).

ext/liquid_c/tokenizer.c
evulse: I'm actually refactoring this entire pull request around this as your right its much better to extend the tokenizer to look forward 3 characters rather than its current 2 so it can see the white space control character and remove any need for the code to ever look backwards.

evulse:
Alright we are passing again. Detection of the whitespace control characters has now been pushed back in to the tokenizer so there is no need to push and pop any existing values.

tobi:
really nice work Mike.


- tobi
CEO Shopify

ext/liquid_c/block.c
```diff
@@ -55,7 +56,15 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
...
-                VALUE str = rb_enc_str_new(token.str, token.length, utf8_encoding);
+                const char *start = token.str, *end = token.str + token.length;
+                const char *token_start = start;
+
+                if(token.lstrip)
+                    token_start = read_while(start, end, rb_isspace);
+
+                VALUE str = rb_enc_str_new(token_start, end - token_start, utf8_encoding);
+                if(token.rstrip)
+                    rb_funcall(str, intern_rstrip, 0);
```
pushrax: What do you think about adding a `reverse_read_while` if the lstrip part is using `read_while` anyway?

pushrax:
Definitely the better solution, nice.

ext/liquid_c/block.c
evulse: Yeah I actually was thinking to do exactly that. I really only stuck to this while refactoring so I could confirm i didn't introduce anything new. Should have some time tonight to pull this out and bring it up to where the lstrip sits.

evulse:
Ok new change from using ruby rstrip is done which makes this pretty final. As such run the final performance benchmark

**Current**

    Calculating -------------------------------------
              parse:     80.921  (± 3.7%) i/s -      4.851k in  60.028478s
        parse & run:     32.196  (± 3.1%) i/s -      1.932k in  60.069380s

**This feature**

    Calculating -------------------------------------
              parse:     80.791  (± 3.7%) i/s -      4.844k in  60.031186s
        parse & run:     32.109  (± 3.1%) i/s -      1.926k in  60.046058s

The results show the performance impact of this feature is very minimal.

evulse:
@pushrax @tobi @dylanahsmith Where to from here? We would love to be able to start using this

ext/liquid_c/tokenizer.c
```diff
@@ -66,20 +67,28 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
     token->type = TOKEN_RAW;
+    token->lstrip = 0;
+    token->rstrip = 0;
 
     while (cursor < last) {
         if (*cursor++ != '{')
             continue;
 
         char c = *cursor++;
+        char w = *cursor++;
```
dylanahsmith: This can read past the end of the string.  E.g. tokenizing `"{{"` results in a token with `"{{\u0000"`.  The `while (cursor < last)` check only allows us to read two characters before having to do another bounds check to see if we can read another byte.

ext/liquid_c/tokenizer.c
```diff
@@ -66,20 +67,28 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
         if (c != '%' && c != '{')
             continue;
-        if (cursor - tokenizer->cursor > 2) {
+        if (cursor - tokenizer->cursor > 3) {
             token->type = TOKEN_RAW;
-            cursor -= 2;
+            cursor -= 3;
+            token->rstrip = (w == '-');
```
dylanahsmith: I would keep the lstrip and rstrip initialization, then replace the other above changes to this function with

```c
if (cursor <= last && *cursor == '-') {
  cursor++;
  token->rstrip = 1;
}
```

ext/liquid_c/tokenizer.c
```diff
@@ -90,10 +99,13 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
                 if (c != '}')
                     continue;
                 token->type = TOKEN_TAG;
+                if(token->str[cursor - tokenizer->cursor - 3] == '-')
```
dylanahsmith: Why not check `cursor[-3]` instead of `token->str[cursor - tokenizer->cursor - 3]`

ext/liquid_c/tokenizer.c
```diff
@@ -90,10 +99,13 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
+                    token->rstrip = tokenizer->lstrip_flag = 1;
                 goto found;
             }
             // unterminated tag
-            cursor = tokenizer->cursor + 2;
+            cursor = tokenizer->cursor + 3;
```
dylanahsmith: Why are you taking three characters for an unterminated tag.  E.g. this means `'{{ab'` would get tokenized into `['{{a', 'b']` instead of `['{{', 'ab']`.  If the intention was to include the whitespace control character in the unterminated token, then I don't see the point in doing so.

ext/liquid_c/tokenizer.c
```diff
@@ -66,20 +67,28 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
         if (c != '%' && c != '{')
             continue;
-        if (cursor - tokenizer->cursor > 2) {
+        if (cursor - tokenizer->cursor > 3) {
             token->type = TOKEN_RAW;
-            cursor -= 2;
+            cursor -= 3;
+            token->rstrip = (w == '-');
+            if(tokenizer->lstrip_flag)
+                token->lstrip = 1;
```
dylanahsmith: You can remove the conditional and just set `token->lstrip  = tokenizer->lstrip_flag;`

ext/liquid_c/tokenizer.c
```diff
@@ -105,14 +117,19 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
             // unterminated variable
-            cursor = tokenizer->cursor + 2;
+            cursor = tokenizer->cursor + 3;
+            tokenizer->lstrip_flag = 0;
             goto found;
         }
     }
     cursor = last + 1;
+    if(tokenizer->lstrip_flag)
+        token->lstrip = 1;
```
dylanahsmith: This can also unconditionally do `token->lstrip = tokenizer->lstrip_flag;`

You should follow that with `tokenizer->lstrip_flag = 0;` as well, like you did above

ext/liquid_c/block.c
```diff
@@ -68,15 +83,15 @@ static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE options)
...
+                VALUE args[2] = {rb_enc_str_new(token.str + 2 + token.lstrip, token.length - 4 - token.lstrip - token.rstrip, utf8_encoding), options};
                 VALUE var = rb_class_new_instance(2, args, cLiquidVariable);
                 rb_ary_push(nodelist, var);
                 rb_ivar_set(self, intern_blank, Qfalse);
                 break;
             }
             case TOKEN_TAG:
             {
-                const char *start = token.str + 2, *end = token.str + token.length - 2;
+                const char *start = token.str + (2 + token.lstrip), *end = token.str + (token.length - 2 - token.rstrip);
```
dylanahsmith: The brackets are unnecessary the way you are using them

dylanahsmith:
There were some corner cases that you missed as noted above.  But overall this is looking very good now.  Nice work

evulse:
@dylanahsmith is this what you were thinking?

ext/liquid_c/tokenizer.c
```diff
@@ -66,20 +67,29 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
+    token->rstrip = 0;
 
     while (cursor < last) {
         if (*cursor++ != '{')
             continue;
 
         char c = *cursor++;
+        if (cursor <= last && *cursor == '-') {
+          cursor++;
+          token->rstrip = 1;
```
dylanahsmith: If the following `if (c != '%' && c != '{')` branch is taken, then this change won't be reversed.

ext/liquid_c/tokenizer.c
```diff
@@ -66,20 +67,29 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
...
+        if (cursor - tokenizer->cursor > 2 + token->rstrip) {
             token->type = TOKEN_RAW;
-            cursor -= 2;
+            cursor -= 2 + token->rstrip;
+            token->lstrip = tokenizer->lstrip_flag;
+            tokenizer->lstrip_flag = 0;
             goto found;
         }
         token->type = TOKEN_INVALID;
+        token->lstrip = token->rstrip;
```
dylanahsmith: This will leave `token->rstrip` set to 1, even though it is just the left whitespace control character that has been found.  Below you only conditionally set `token->rstrip = 1`, so left whitespace control character will always result in the rstrip flag being set on the token.

evulse:
I think just shifting this removes this issue. So flowing through this previously this would have allowed `{.-` where . being any character. However i think shifting this only allows `{%-` and `{{-` as the tag. So the only flows are

- These 3 characters are at the end of a string and become a raw token
- These 3 characters are at the start and become a token or variable and run their own loops
- This is the last token and as such there will be no other tag following that will trigger the lstrip

evulse:
Is there anything else needed for this to progress?

dylanahsmith:
It looks like https://github.com/Shopify/liquid-c/pull/30/files#r70563502 still hasn't been addressed.  That was my last concern.

evulse:
Ok i've added a reset after this line so it will only be set if it is found.

evulse:
Ok this and the ruby version look like they are ready to go. What do we need to do to get this moving.

PR #31, opened 2016-08-28
Fix "comparison of integers of different signs"

kainjow:
Got this error with latest Xcode installed on 10.11.5:

```
tokenizer.c:84:40: error: comparison of integers of different signs: 'int' and 'unsigned int' [-Werror,-Wsign-compare]
        if (cursor - tokenizer->cursor > 2 + token->rstrip) {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/ext/liquid_c/tokenizer.c
+++ b/ext/liquid_c/tokenizer.c
@@ -81,7 +81,7 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
           cursor++;
           token->rstrip = 1;
         }
-        if (cursor - tokenizer->cursor > 2 + token->rstrip) {
+        if (cursor - tokenizer->cursor > (intptr_t)(2 + token->rstrip)) {
             token->type = TOKEN_RAW;
             cursor -= 2 + token->rstrip;
             token->lstrip = tokenizer->lstrip_flag;
--- a/ext/liquid_c/tokenizer.c
+++ b/ext/liquid_c/tokenizer.c
@@ -81,7 +81,7 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
           cursor++;
           token->rstrip = 1;
         }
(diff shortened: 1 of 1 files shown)


ext/liquid_c/tokenizer.c
```diff
@@ -81,7 +81,7 @@ void tokenizer_next(tokenizer_t *tokenizer, token_t *token)
           cursor++;
           token->rstrip = 1;
         }
-        if (cursor - tokenizer->cursor > 2 + token->rstrip) {
+        if (cursor - tokenizer->cursor > (intptr_t)(2 + token->rstrip)) {
```
fw42: why cast to `intptr_t`? looking at your error message, I would have expected a cast to `int` or `unsigned int`

kainjow: `cursor` and `tokenizer->cursor` are of type `const char*` so a pointer type should be used. As for "int" and "unsigned int", it may have been compiling for 32-bit, as Ruby still seems to be a fat binary on OS X:

> $ file /usr/bin/ruby
> /usr/bin/ruby: Mach-O universal binary with 2 architectures
> /usr/bin/ruby (for architecture x86_64):	Mach-O 64-bit executable x86_64
/usr/bin/ruby (for architecture i386):	Mach-O executable i386

Technically, `ptrdiff_t` should probably be used, but I don't know if that really matters.

fw42: cool, thanks for the explanation

fw42:
@pushrax?

ext/liquid_c/tokenizer.c
pushrax: I agree that `ptrdiff_t` is correct here, if we're fixing this specifically may as well do it right :)

kainjow: Code has been updated to use ptrdiff_t

pushrax:
Thanks!

PR #33, opened 2017-06-20
Update Travis for supported Rubies

bdewater:
Especially with Ruby 2.4 where the C API breaks because of the Fixnum/Bignum unification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: ruby
 
 rvm:
-  - 2.0
-  - 2.1
   - 2.2
+  - 2.3
+  - 2.4.1
   - ruby-head
 
 sudo: false


dylanahsmith:
It doesn't look like the Fixnum/Bignum unification has affected the ruby C API from what I can tell.

However, thanks for fixing the tests, since it would otherwise fail on ruby 2.0 that is no longer supported

bdewater:
"All C extensions which touch the Fixnum or Bignum class need to be fixed" from https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/ but for this gem it wasn't a problem 😄

dylanahsmith:
Right, references to rb_cFixnum or rb_cBignum, but macros like INT2FIX haven't been removed.  Either way, it is nice to have CI catch any problems

PR #34, opened 2017-06-20
Refactor undefined default value for hash lookup

dylanahsmith:
## Problem

I noticed a weird line when looking for references to fixnum in the codebase

```ruby
VALUE undefined = FIXNUM_P(-1);
```

which seems very weird to me, since `FIXNUM_P` is meant to be used in a predicates (e.g. `if (FIXNUM_P(obj)) {`) and not a ruby `VALUE`.  I think maybe the code intended to do `INT2FIX(-1)` instead.  Instead, I think it is unintentionally relying on implementation details to continue working.

`FIXNUM_P(-1)` happens to be equivalent to `INT2FIX(0)` in the current implementation of MRI ruby, which only works since it isn't a value in the expresssion hash that this is used as a default value for.

## Solution

Use `Qundef` instead, since it is intended to be used as a value that can't be specified from ruby code.

### DIFF
--- a/ext/liquid_c/parser.c
+++ b/ext/liquid_c/parser.c
@@ -120,9 +120,8 @@ static VALUE parse_variable(parser_t *p)
     }
 
     if (RARRAY_LEN(lookups) == 0) {
-        VALUE undefined = FIXNUM_P(-1);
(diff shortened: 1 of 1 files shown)


dylanahsmith:
Closes #38

PR #38, opened 2018-01-11
Fix Hash lookup with default value in parser.c

eregon:
Hello,
We noticed this issue when trying this C extension on TruffleRuby.
The `FIXNUM_P(-1)` part seems rather odd as `FIXNUM_P()` should take a `VALUE` (an address).
`-1` happens to be the address of Fixnum `-1` on MRI, so `FIXNUM_P(-1)` returns `(VALUE) 1` which is Ruby Fixnum `0` and accidentaly works because there is no Fixnum `0` in the values of https://github.com/Shopify/liquid/blob/64fca66ef5dfd7cf1acef0a7b8cdd825756eb681/lib/liquid/expression.rb#L16-L22.

* The default value should be Qundef, not an arbitrary ((VALUE) 1).
* Qundef is not available in Ruby so it can safely be differentiated
  from Ruby-level values.

### DIFF
--- a/ext/liquid_c/parser.c
+++ b/ext/liquid_c/parser.c
@@ -120,9 +120,8 @@ static VALUE parse_variable(parser_t *p)
     }
 
     if (RARRAY_LEN(lookups) == 0) {
-        VALUE undefined = FIXNUM_P(-1);
-        VALUE literal = rb_hash_lookup2(vLiquidExpressionLiterals, name, undefined);
-        if (literal != undefined) return literal;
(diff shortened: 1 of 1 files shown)


nirvdrum:
This looks like a dupe of #34.

dylanahsmith:
Sorry for forgetting to merge #34.

eregon:
Indeed, I should have looked opened PRs.
Thanks for getting the fix in master!

Is there any plan to make a release?

PR #40, opened 2018-11-08
CI: test against supported Ruby versions

DirtyF:
- Ruby 2.2 is EOL since march 2018
 - current stable version is 2.5
 - ruby-head is now a 2.6@dev

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: ruby
 
 rvm:
-  - 2.2
   - 2.3
-  - 2.4.1
+  - 2.4
+  - 2.5
   - ruby-head
 
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 
 rvm:
-  - 2.3
+  - 2.2
   - 2.5
   - ruby-head
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: ruby
 
 rvm:
   - 2.3
-  - 2.4
   - 2.5
   - ruby-head
 


.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: ruby
 
 rvm:
-  - 2.2
   - 2.3
-  - 2.4.1
+  - 2.4
```
dylanahsmith: I don't think we need to test all supported versions of ruby.  I think that is only necessary if we have ruby version specific code.  I think we only need to test the latest ruby version and the minimum supported ruby version.

.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: ruby
 
 rvm:
-  - 2.2
```
dylanahsmith: This does nothing to make the minimum ruby version clear to users of this library.  That is what the `required_ruby_version` in the gemspec does.  Removing the minimum version from CI just means we are more likely to accidentally break a ruby version and makes the actual minimum ruby version implicit.

DirtyF: added require_ruby_version 2.3 in #41

.travis.yml
DirtyF: 2.4 testing is now removed

.travis.yml
dylanahsmith: But why drop support for older ruby versions without a reason?  The liquid gem supports ruby `>= 2.1.0`.  Also, the last time we bumped the ruby requirement based on what was supported by ruby upstream, it caused problems for users using the version of ruby already installed on their system.

DirtyF: Mainly for security and performance. I didn't know that liquid gem still supported `>= 2.1`. In my experience current systems support 2.3 out of the box, I thought it was a safe option. I don't know about your user base.
Ruby 2.1 and 2.2 are not supported anymore, and Ruby 2.3 going EOL in march, so the question is when do you decide to bump? I tend to follow Ruby release cycle.

dylanahsmith: Oh, maybe MacOS is keeping ruby more up to date now. That didn't seem to be the case before (https://github.com/Shopify/liquid/pull/832#issuecomment-264535801).

We aren't preventing users from upgrading for security or performance reasons.  However, when we bump the ruby version, it normally is based on the ruby release cycle.  I just don't see the benefit of bumping the minimum version for no reason.

We will bump the ruby version when we want to take advantage of a ruby feature in a newer version.  For instance, we will want to take advantage of symbol GC from ruby 2.2 in the foreseeable future.

PR #42, opened 2019-03-08
Skip allocating empty raw tokens

pushrax:
Tiny change to reduce object allocation. Avoids allocating empty strings between tags with whitespace control `-}}   {{-`.

The ruby-head build is broken in general, tests do pass on 2.x.


dylanahsmith:
> The ruby-head build is broken in general, tests do pass on 2.x.

Should we [allow_failures](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail) until that is fixed?

pushrax:
Fixed ruby-head, was just an overly tight dependency on Bundler.

PR #45, opened 2019-04-12
Implicit variable scoping

fw42:
Counterpart to https://github.com/Shopify/liquid/pull/1090.

### DIFF
--- a/ext/liquid_c/context.c
+++ b/ext/liquid_c/context.c
@@ -5,7 +5,7 @@
 static VALUE cLiquidVariableLookup, cLiquidUndefinedVariable;
 ID id_aset, id_call, id_to_liquid, id_set_context;
 static ID id_evaluate, id_has_key, id_aref;
-static ID id_ivar_scopes, id_ivar_environments, id_ivar_strict_variables;
+static ID id_ivar_scope, id_ivar_environments, id_ivar_strict_variables;
 
 VALUE context_evaluate(VALUE self, VALUE expression)
 {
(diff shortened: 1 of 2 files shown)


pushrax:
If you update the Gemfile to point to right Liquid branch, we should be able to see tests pass.

ext/liquid_c/context.c
```diff
@@ -41,25 +41,13 @@ VALUE context_find_variable(VALUE self, VALUE key, VALUE raise_on_not_found)
...
-                scope = this_scope;
-                goto variable_found;
-            }
-        } else if (RTEST(rb_funcall(this_scope, id_has_key, 1, key))) {
-            // Slow path: It is valid to pass a non-hash value to Liquid as a
-            // scope if it supports #key? and #[]
-            variable = rb_funcall(this_scope, id_aref, 1, key);
-            goto variable_found;
-        }
+    variable = rb_hash_lookup2(scope, key, Qnil);
```
pushrax: To make this work identically to `#key?` as used here https://github.com/Shopify/liquid/pull/1090/files#diff-d97dbf3bc11085a7aa13ee71b9078b85R146 it should use `Qundef`. If the scope explicitly has the variable set as nil it should not be returned here.

fw42: Yea good catch, I did this wrong on the Ruby PR as well I guess.

pushrax: Ugh my comment was backwards from what I was thinking... if the scope explicitly has the variable set as nil it _should_ be returned, and you have this right in the Ruby PR. That's how it worked before your change as well. The C PR is different though, it will not return the nil value.

fw42:
Tests are passing now

PR #51, opened 2019-11-05
Honor pre-trim for tags

XrXr:
Not the fastest way to do it -- I think doing it in the tokenizer is
better. Let me know if this is acceptable.

### DIFF
--- a/test/unit/block_test.rb
+++ b/test/unit/block_test.rb
@@ -8,4 +8,12 @@ class BlockTest < MiniTest::Test
     template = Liquid::Template.parse("{{ -}} foo {{- }}")
     assert_equal 3, template.root.nodelist.size
   end
+
+  def test_pre_trim
+    template = Liquid::Template.parse("\n{%- raw %}{% endraw %}")
+    assert_equal "", template.render
+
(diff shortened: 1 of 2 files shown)


pushrax:
I'm not sure why https://github.com/Shopify/liquid-c/blob/76b1b8ae5268ecdab4441e4701eafe83b9e0c804/ext/liquid_c/tokenizer.c#L120-L129 isn't making this work already.

XrXr:
> I'm not sure why

Turns out, it's an off-by-one that I missed!

test/unit/block_test.rb
```diff
@@ -8,4 +8,12 @@ def test_no_allocation_of_trimmed_strings
...
     assert_equal 3, template.root.nodelist.size
   end
+
+  def test_pre_trim
+    template = Liquid::Template.parse("\n{%- raw %}{% endraw %}")
+    assert_equal "", template.render
+
+    template = Liquid::Template.parse("\n{%- if true %}{% endif %}")
+    assert_equal "", template.render
+  end
```
dylanahsmith: This should really be a part of the liquid integration tests, since it isn't liquid-c specific

PR #52, opened 2019-12-19
Fix liquid tag nested in outer block

dylanahsmith:
The bug https://github.com/Shopify/liquid/pull/1215 is fixing is also present in the C code, so this PR makes those corresponding changes to the C code.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -13,14 +13,27 @@ static ID
     intern_registered_tags,
     intern_parse,
     intern_square_brackets,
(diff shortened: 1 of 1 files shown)


dylanahsmith:
The ruby-head CI failure

```
  1) Failure:
ContextTest#test_evaluating_a_variable_entirely_within_c [/Users/dylansmith/src/liquid-c/test/unit/context_test.rb:51]:
Expected: 0
  Actual: 2
```

is also on liquid-c master, so isn't related to this PR, although should be investigated.

PR #53, opened 2019-12-19
Fix tracepoint test on ruby master

dylanahsmith:
As mentioned in https://github.com/Shopify/liquid-c/pull/52#issuecomment-567293796, CI started failing on ruby master.

The failing test is counting ruby calls using a tracepoint and started picking up a couple of calls to TracePoint.trace and TracePoint#disable.  We were already ignoring disable C calls, so I changed the code to ignore both ruby and C tracepoint calls, since they aren't relevant to the code under test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 
 rvm:
   - 2.4
-  - 2.6
+  - 2.7
   - ruby-head
 
 sudo: false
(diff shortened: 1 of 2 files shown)


dylanahsmith:
> So I'm going to just ignore ruby-head failures on CI for this repo.

Tests are fixed in the liquid repo, so I've removed the allowing of test failures on ruby-head.  I've also made CI test against ruby 2.7 now that it has been released

PR #54, opened 2020-02-03
Temporarily allow opt-in bug compatible whitespace trimming

dylanahsmith:
to allow us to test the impact of the https://github.com/Shopify/liquid-c/pull/51 fix in an opt-in way before fully rolling it out.


XrXr:
It might be nice to mention that this "feature" is not here to stay in the code.

PR #58, opened 2020-09-08
Add and use Liquid::C::BlockBody for the block body

dylanahsmith:
## Problem

Currently liquid-c is mostly used to speed up parsing (evaluating expressions is the exception).  However, there is a big opportunity to reduce the overhead of both parsing and rendering by moving to a C representation for block bodies, since we can represent it as a set of instructions that can be efficiently executed directly from C rather than indirectly through another VM and would also allow us to serialize and deserialize it to avoid parsing overhead for pre-compiled templates. The block body is the ideal starting place, since it can encapsulate everything it contains, allowing builtin functionality to be compiled directly into the block bodies instructions.

@macournoyer created a proof of concept liquid VM as part of hack days, which showed significant potential improvements from a liquid VM.  Even after modifying it to use a Liquid::Context rather than a simple hash for the context, it was still showing a 4.4x improvement for deserializing and rendering the liquid VM code compared to parsing and rendering liquid using liquid-c.  Loading a marshal dump of the Liquid::Template instead of parsing liquid only made it 4.0x slower than liquid VM code, which seems to indicate that we should serialize to a format that can (at least mostly) be executed without deserialization.

## Solution

This pull request depends on https://github.com/Shopify/liquid/pull/1289 to provide Block#new_body and Document#new_body to override in order to return a Liquid::C::BlockBody when liquid-c is enabled.  These two methods also have access to the parse context, so we still don't need to support profiling for liquid-c's block body.

The code implementing the Liquid::BlockBody#parse override was refactored to implement Liquid::C::BlockBody#parse, so we can build the Liquid::C::BlockBody directly without the overhead of creating a `Liquid::BlockBody#nodelist` array.  This also allows us to avoid copying the raw template strings, which was previously needed to create individual ruby strings, since we can instead just write the output directly from slices of a shared string that contains all the raw template strings.

To keep the scope of this PR more minimal, I haven't included block body serialization and have not added compilation of variables or tags.  So the VM currently only has 3 instructions: OP_LEAVE, OP_WRITE_RAW and OP_WRITE_NODE.

The VM keeps track of both an instruction pointer and a constant pointer, which are incremented as instructions or constants are used, so the instructions will be easy to used directly from a serialized state and we won't have to worry about complexities of encoding a constant references (e.g. a byte constant index is too limiting and a multi-byte index can introduce either extra alignment or decoding concerns).  We can always iterate on the VM design or even compile directly to machine code in the future.

To make this easier to adopt Liquid::C::BlockBody, I added Liquid::C::BlockBody#nodelist, although I will also be refactoring hot code paths using that in Shopify to not depend on it for performance reasons.  ~~As part of our [storefront rewrite](https://engineering.shopify.com/blogs/engineering/how-shopify-reduced-storefront-response-times-rewrite) we have some code to debug differences in the rendered output, which is quite coupled to the liquid parse tree node ruby objects, but doesn't seem to be performance sensitive, so I added a `:disable_liquid_c_nodes` parse option so this can be used without disabling liquid-c globally (which wouldn't be thread-safe).~~

## Benchmark

On master:

```
              parse:    146.973  (± 4.1%) i/s -      1.470k in  10.020119s
             render:    135.523  (± 3.0%) i/s -      1.365k in  10.083170s
     parse & render:     64.774  (± 3.1%) i/s -    648.000  in  10.010060s
```

On this PR's branch:

```
              parse:    145.177  (± 2.1%) i/s -      1.456k in  10.032703s
             render:    157.193  (± 5.1%) i/s -      1.575k in  10.048723s
     parse & render:     68.667  (± 2.9%) i/s -    690.000  in  10.060160s
```

## Check List

- [x] Get https://github.com/Shopify/liquid/pull/1289 reviewed and merged
- [x] Get https://github.com/Shopify/liquid/pull/1285 approved and merged (or rebase code to not depend on it)
- [x] Remove the "Temporarily use liquid branch until it has been merged" commit

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -23,15 +23,20 @@ typedef struct tag_markup {
     VALUE markup;
 } tag_markup_t;
 
+typedef struct parse_context {
+    tokenizer_t *tokenizer;
+    VALUE tokenizer_obj;
+    VALUE ruby_obj;
+} parse_context_t;
+
 static int is_id(int c)
 {
     return rb_isalnum(c) || c == '_';
 }
 
-static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_context)
+static tag_markup_t internal_block_parse(VALUE self, const parse_context_t *parse_context)
 {
-    tokenizer_t *tokenizer;
-    Tokenizer_Get_Struct(tokens, tokenizer);
+    tokenizer_t *tokenizer = parse_context->tokenizer;
 
     token_t token;
     VALUE tags = Qnil;
@@ -41,7 +46,7 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
     while (true) {
         int token_start_line_number = tokenizer->line_number;
         if (token_start_line_number != 0) {
-            rb_funcall(parse_context, intern_set_line_number, 1, UINT2NUM(token_start_line_number));
+            rb_funcall(parse_context->ruby_obj, intern_set_line_number, 1, UINT2NUM(token_start_line_number));
         }
         tokenizer_next(tokenizer, &token);
 
@@ -56,7 +61,7 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
                 ID raise_method_id = intern_raise_missing_variable_terminator;
                 if (token.str_full[1] == '%') raise_method_id = intern_raise_missing_tag_terminator;
 
-                rb_funcall(self, raise_method_id, 2, str, parse_context);
+                rb_funcall(self, raise_method_id, 2, str, parse_context->ruby_obj);
                 return unknown_tag;
             }
             case TOKEN_RAW:
@@ -92,7 +97,7 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
             }
             case TOKEN_VARIABLE:
             {
-                VALUE args[2] = {rb_enc_str_new(token.str_trimmed, token.len_trimmed, utf8_encoding), parse_context};
+                VALUE args[2] = {rb_enc_str_new(token.str_trimmed, token.len_trimmed, utf8_encoding), parse_context->ruby_obj};
                 VALUE var = rb_class_new_instance(2, args, cLiquidVariable);
                 rb_ary_push(nodelist, var);
                 rb_ivar_set(self, intern_blank, Qfalse);
@@ -123,10 +128,10 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
 
                     tokenizer_t saved_tokenizer = *tokenizer;
                     tokenizer_setup_for_liquid_tag(tokenizer, markup_start, end, line_number);
-                    unknown_tag = internal_block_parse(self, tokens, parse_context);
+                    unknown_tag = internal_block_parse(self, parse_context);
                     *tokenizer = saved_tokenizer;
                     if (unknown_tag.name != Qnil) {
-                        rb_funcall(cLiquidBlockBody, intern_unknown_tag_in_liquid_tag, 2, unknown_tag.name, parse_context);
+                        rb_funcall(cLiquidBlockBody, intern_unknown_tag_in_liquid_tag, 2, unknown_tag.name, parse_context->ruby_obj);
                         return unknown_tag;
                     }
                     break;
@@ -148,7 +153,8 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
                     return unknown_tag;
                 }
 
-                VALUE new_tag = rb_funcall(tag_class, intern_parse, 4, tag_name, markup, tokens, parse_context);
+                VALUE new_tag = rb_funcall(tag_class, intern_parse, 4,
+                        tag_name, markup, parse_context->tokenizer_obj, parse_context->ruby_obj);
 
                 if (rb_ivar_get(self, intern_blank) == Qtrue && !RTEST(rb_funcall(new_tag, intern_is_blank, 0)))
                     rb_ivar_set(self, intern_blank, Qfalse);
@@ -161,9 +167,15 @@ static tag_markup_t internal_block_parse(VALUE self, VALUE tokens, VALUE parse_c
     return unknown_tag;
 }
 
-static VALUE rb_block_parse(VALUE self, VALUE tokens, VALUE parse_context)
+static VALUE rb_block_parse(VALUE self, VALUE tokenizer_obj, VALUE parse_context_obj)
 {
-    tag_markup_t unknown_tag = internal_block_parse(self, tokens, parse_context);
+    parse_context_t parse_context = {
+        .tokenizer_obj = tokenizer_obj,
+        .ruby_obj = parse_context_obj,
+    };
+    Tokenizer_Get_Struct(tokenizer_obj, parse_context.tokenizer);
+
+    tag_markup_t unknown_tag = internal_block_parse(self, &parse_context);
     return rb_yield_values(2, unknown_tag.name, unknown_tag.markup);
 }
 
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -1,22 +1,20 @@
 #include "liquid.h"
+#include "block.h"
 #include "tokenizer.h"
 #include "stringutil.h"
+#include "vm.h"
 #include <stdio.h>
 
 static ID
     intern_raise_missing_variable_terminator,
     intern_raise_missing_tag_terminator,
-    intern_nodelist,
-    intern_blank,
     intern_is_blank,
-    intern_clear,
-    intern_registered_tags,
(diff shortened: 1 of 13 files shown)


ext/liquid_c/vm.c
```diff
@@ -0,0 +1,75 @@
...
+    VALUE interrupts = rb_ivar_get(context, id_ivar_interrupts);
+    Check_Type(interrupts, T_ARRAY);
+
+    while (true) {
+        switch (*ip++) {
+            case OP_LEAVE:
+                return;
+            case OP_WRITE_RAW:
+            {
+                const char *text = (const char *)*const_ptr++;
```
macournoyer, line 42: Very nice and simple way to avoid index overflow 👍 

However, it will break once we introduce loops (`for`, etc.)

lib/liquid/c.rb
```diff
@@ -14,22 +20,38 @@ def self.new(source, line_numbers = false, line_number: nil, for_liquid_tag: fal
...
-    else
-      ruby_parse(tokens, options) { |t, m| yield t, m }
+[Liquid::Block, Liquid::Document].each do |klass|
+  klass.class_eval do
+    private
+
+    alias_method :ruby_new_body, :new_body
+
+    def new_body
+      if parse_context.disable_liquid_c_nodes || parse_context[:profile]
```
macournoyer, line 39: Does profiling disables this only because instrumentation is missing from `C::BlockBody` at the moment?

macournoyer (approved):
Impressive work 👏 

A few comments, no blockers, as long as we can change the bytecode later.

ext/liquid_c/vm.c
dylanahsmith: I do plan on revisiting the moving constant pointer when adding jumps.  It could be implemented by have the jump target be a pair of instruction and constant pointer offsets, so there is a trade-off between jump overhead and constant lookup overhead.

lib/liquid/c.rb
dylanahsmith: Liquid's template render profiling support already wasn't supported in liquid-c.  I just chose to keep it that way.  Note that this is different from stackprof, which we typically use for profiling at Shopify.  I would like to revisit profiling in the future, but not in a way that ties us to calling a `render_node` ruby method that requires a node object to be passed to it.

ext/liquid_c/block.c
```diff
@@ -1,59 +1,142 @@
...
 
+static inline void block_body_add_leave(block_body_t *body)
+{
+    block_body_write_opcode(body, OP_LEAVE);
+}
+
+static void block_body_add_write_raw(block_body_t *body, const char *string, size_t size)
+{
+    block_body_write_opcode(body, OP_WRITE_RAW);
+    size_t slice[2] = { (size_t)string, size };
```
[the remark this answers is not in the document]
dylanahsmith: There will already be constants that we won't use directly from the serialized format, such as uncompiled tags or expression constants (e.g. strings, big integers, constant ranges).  In the serialized format, there will end up being a split between these constants to reduce deserialization overhead, so that at least the data for constants like raw strings can be used without deserializing.

I don't think it would be that much overhead for resolving the pointers to raw strings at deserialization, but as mentioned in https://github.com/Shopify/liquid-c/pull/58#discussion_r485247888, I do plan on revisiting the decision of using a moving constant pointer.

ext/liquid_c/vm.c
macournoyer: Sounds good. First time seeing this approach...

But when you'll add a pointer offset, I think it will be equivalent to the traditional `constants[i]`, and you'll end up needing multi-byte index too.

felix-d:
> As part of our storefront rewrite we have some code to debug differences in the rendered output, which is quite coupled to the liquid parse tree node ruby objects, but doesn't seem to be performance sensitive, so I added a :disable_liquid_c_nodes parse option so this can be used without disabling liquid-c globally (which wouldn't be thread-safe).

We don't use liquid tracking anymore. This can be 🔥

dylanahsmith:
Oh right, I removed the use of `:disable_liquid_c_nodes` in the storefront renderer PR, but didn't completely remove the option here.  I've done that now.

ext/liquid_c/block.c
```diff
@@ -123,65 +210,199 @@ static VALUE internal_block_parse(VALUE self, VALUE tokens, VALUE parse_context,
...
+    BlockBody_Get_Struct(self, body);
+
+    ensure_not_parsing(body);
+
+    // Use rb_attr_get insteaad of rb_ivar_get to avoid an instance variable not
+    // initialized warning.
+    VALUE nodelist = rb_attr_get(self, intern_ivar_nodelist);
+    if (nodelist != Qnil)
+        return nodelist;
+    nodelist = rb_ary_new_capa(body->instructions.size / sizeof(VALUE));
```
pushrax, line 353: Since this is just the capacity it doesn't matter, but it looks like this is including an extra index for the OP_LEAVE.

ext/liquid_c/vm.c
```diff
@@ -0,0 +1,75 @@
...
+ID id_raise_limits_reached;
+
+void liquid_vm_render(block_body_t *body, VALUE context, VALUE output)
+{
+    Check_Type(output, T_STRING);
+
+    VALUE resource_limits = rb_ivar_get(context, id_ivar_resource_limits);
+    rb_funcall(resource_limits, id_increment_render_score, 1, INT2NUM(body->render_score));
+
+    bool is_captured = rb_ivar_get(resource_limits, id_ivar_last_capture_length) != Qnil;;
```
pushrax, line 22: nit: extra ;

ext/liquid_c/vm.c
```diff
@@ -0,0 +1,75 @@
...
+        if (render_length_limit_num != Qnil)
+            render_length_limit = NUM2LONG(render_length_limit_num);
+    }
+
+    const size_t *const_ptr = (const size_t *)body->constants.data;
+    const uint8_t *ip = body->instructions.data;
+    VALUE interrupts = rb_ivar_get(context, id_ivar_interrupts);
+    Check_Type(interrupts, T_ARRAY);
+
+    while (true) {
```
pushrax, line 36: Would it make sense to do some tracking for how many bytes of the instruction buffer have been consumed, to gracefully handle a corrupted buffer with a missing OP_LEAVE? Perhaps by computing `last_ip` at the start and doing a single comparison per iteration.

pushrax (approved):
Very nice, this was easy to follow and looks quite efficient. Also just had some minor non-blocking comments.

ext/liquid_c/block.c
```diff
@@ -1,59 +1,142 @@
...
+}
+
+static VALUE block_body_allocate(VALUE klass)
+{
+    block_body_t *body;
+
+    VALUE obj = TypedData_Make_Struct(klass, block_body_t, &block_body_data_type, body);
+    body->instructions = c_buffer_allocate(8);
+    block_body_add_leave(body);
+    body->constants = c_buffer_allocate(8 * sizeof(VALUE));
```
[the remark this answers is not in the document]
dylanahsmith: It was chosen fairly arbitrarily based on the current situation where we could end up with a lot of small block bodies for things like `if` tags.  I tried to keep it as a multiple of 2, since that seemed like it would make it more likely that the memory allocator will re-use buffer allocations for new or expanding buffers and because it could use whole pages for larger memory allocations.  However, I haven't tested any of these assumptions yet.

We will probably want to revisit the default Liquid::Document's block body as we start compiling more functionality into VM instructions, since then we will be able to compile builtin block tags into their parent block body.

ext/liquid_c/block.c
```diff
@@ -123,65 +210,199 @@ static VALUE internal_block_parse(VALUE self, VALUE tokens, VALUE parse_context,
...
+            case OP_LEAVE:
+                goto loop_break;
+            case OP_WRITE_RAW:
+            {
+                const_ptr++;
+                size_t *size_ptr = const_ptr++;
+                if (*size_ptr) {
+                    *size_ptr = 0; // effectively a no-op
+                    body->render_score--;
+                }
```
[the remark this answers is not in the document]
dylanahsmith: As mentioned in https://github.com/Shopify/liquid-c/pull/59#discussion_r490293174

> I'm not concerned about further optimizing this edge case. Primarily these blank strings are handled in this way to minimize impact on code that isn't relying on this feature while keeping backwards compatibility,

Specifically, once we have jump instructions for control flow tags, we would also be forced to update all the affected jump offsets.  That will be easier to introduce without having to deal with jump offset relocation, but afterwards an extra pass could be added that is more clearly beneficial, since we could replace pessimistically large forward jump instructions (e.g. a decoded 24-bit offset or aligned 32-bit offset) with smaller jump instructions (e.g. a byte offset) where possible.

ext/liquid_c/vm.c
dylanahsmith: I would prefer to not silently try to handle a corrupt buffer, since that seems likely to lead to more problems.  However, I could add an assertion to help with debugging a problem like that.

pushrax: The idea would be to do something like an `rb_bug` in that case, yes.

PR #60, opened 2020-09-24
Implement Liquid::C::Expression to optimize expression evaluation

dylanahsmith:
~~Depends on https://github.com/Shopify/liquid-c/pull/59 and https://github.com/Shopify/liquid/pull/1300~~

## Problem

In https://github.com/Shopify/liquid-c/pull/59, expressions were compiled to ruby constants or lookup ruby objects (Liquid::VariableLookup & Liquid::RangeLookup) that are evaluated with `context_evaluate`.  This was done in that PR for simplicity, but they should really be compiled to more granular VM instructions for rendering performance and for serializability (for future caching of VM compiled templates).

## Solution

I replaced the previous code for strict parsing expressions and variable lookups to instead emit VM code, which is now compiled directly into Liquid::C::BlockBody objects when compiling variables.  Expressions are also often parsed and evaluated independent of Liquid::Variable in tags, so I introduced a `Liquid::C::Expression` object that gets returned from `Liquid::Expression.parse` if the expression includes a variable lookup and strict parsing of it succeeds.  I ended up doing both of these in the same commit, since I also leverage the `Liquid::C::Expression` object for filter keyword argument parsing to handle automatic cleanup on a parse exception.

Note that a couple of direct uses of `Liquid::Variable.new` (in assign and echo tags) no longer leverage liquid-c for parsing, which slightly impacts parse speed.  I'll embed those tags in the liquid VM code in a follow-up PR rather than introducing a Liquid::C::Variable class.  This PR significantly speeds up rendering, as shown in the benchmarks, which more than makes up for this parsing regression.

Since `Liquid::Variable.new` is no longer monkey patched and `Liquid::Expression.parse` no longer turns a comparable object, I had to update the liquid-c variable unit tests so they are tested through normal template parsing and rendering like an integration test.  I also stopped testing against the liquid gem's variable unit tests for the same reasons.

To help with detecting and debugging VM stack underflows or overflows, I added some assertions that are enabled by default for CI and gem development but are disabled by default for gem installation.

## Benchmarks

Before this PR (on the https://github.com/Shopify/liquid-c/pull/59 branch)

```
              parse:    167.388  (± 2.4%) i/s -      1.680k in  10.043356s
             render:    173.196  (± 3.5%) i/s -      1.734k in  10.022222s
     parse & render:     80.409  (± 1.2%) i/s -    808.000  in  10.050380s
```

after

```
              parse:    162.660  (± 2.5%) i/s -      1.632k in  10.040800s
             render:    231.388  (± 4.3%) i/s -      2.323k in  10.059088s
     parse & render:     86.735  (± 1.2%) i/s -    872.000  in  10.056157s
```

### DIFF
--- a/ext/liquid_c/expression.c
+++ b/ext/liquid_c/expression.c
@@ -1,0 +1,97 @@
+#include "liquid.h"
+#include "vm_assembler.h"
+#include "parser.h"
+#include "vm.h"
+#include "expression.h"
+
+VALUE cLiquidCExpression;
+
+static void expression_mark(void *ptr)
+{
+    expression_t *expression = ptr;
+    vm_assembler_gc_mark(&expression->code);
+}
+
+static void expression_free(void *ptr)
+{
+    expression_t *expression = ptr;
+    vm_assembler_free(&expression->code);
+    xfree(expression);
+}
+
+static size_t expression_memsize(const void *ptr)
+{
+    const expression_t *expression = ptr;
+    return sizeof(expression_t) + vm_assembler_alloc_memsize(&expression->code);
+}
+
+const rb_data_type_t expression_data_type = {
+    "liquid_expression",
+    { expression_mark, expression_free, expression_memsize, },
+    NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+VALUE expression_new(expression_t **expression_ptr)
+{
+    expression_t *expression;
+    VALUE obj = TypedData_Make_Struct(cLiquidCExpression, expression_t, &expression_data_type, expression);
+    *expression_ptr = expression;
+    vm_assembler_init(&expression->code);
+    return obj;
+}
+
+static VALUE internal_expression_parse(parser_t *p)
+{
+    if (p->cur.type == TOKEN_EOS)
+        return Qnil;
+
+    // Avoid allocating an expression object just to wrap a constant
+    VALUE const_obj = try_parse_constant_expression(p);
+    if (const_obj != Qundef)
+        return const_obj;
+
+    expression_t *expression;
+    VALUE expr_obj = expression_new(&expression);
+
+    parse_and_compile_expression(p, &expression->code);
+    vm_assembler_add_leave(&expression->code);
+
+    return expr_obj;
+}
+
+static VALUE expression_strict_parse(VALUE klass, VALUE markup)
+{
+    StringValue(markup);
+    char *start = RSTRING_PTR(markup);
+
+    parser_t p;
+    init_parser(&p, start, start + RSTRING_LEN(markup));
+    VALUE expr_obj = internal_expression_parse(&p);
+
+    if (p.cur.type != TOKEN_EOS)
+        rb_enc_raise(utf8_encoding, cLiquidSyntaxError, "[:%s] is not a valid expression", symbol_names[p.cur.type]);
+
+    return expr_obj;
+}
+
+#define Expression_Get_Struct(obj, sval) TypedData_Get_Struct(obj, expression_t, &expression_data_type, sval)
+
+static VALUE expression_evaluate(VALUE self, VALUE context)
+{
+    expression_t *expression;
+    Expression_Get_Struct(self, expression);
+    return liquid_vm_evaluate(context, &expression->code);
+}
+
+VALUE internal_expression_evaluate(expression_t *expression, VALUE context)
+{
+    return liquid_vm_evaluate(context, &expression->code);
+}
+
+void init_liquid_expression()
+{
+    cLiquidCExpression = rb_define_class_under(mLiquidC, "Expression", rb_cObject);
+    rb_undef_alloc_func(cLiquidCExpression);
+    rb_define_singleton_method(cLiquidCExpression, "strict_parse", expression_strict_parse, 1);
+    rb_define_method(cLiquidCExpression, "evaluate", expression_evaluate, 1);
+}
--- a/ext/liquid_c/parser.c
+++ b/ext/liquid_c/parser.c
@@ -2,8 +2,8 @@
 #include "parser.h"
 #include "lexer.h"
 
-static VALUE cLiquidRangeLookup, cLiquidVariableLookup, cRange, vLiquidExpressionLiterals;
-static ID idToI, idEvaluate;
+static VALUE empty_string;
+static ID id_to_i, idEvaluate;
 
 void init_parser(parser_t *p, const char *str, const char *end)
 {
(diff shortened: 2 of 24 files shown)


test/unit/expression_test.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  def test_lookup_const_key
+    context = Liquid::Context.new({"obj" => { "prop" => "some value" }})
+
+    expr = Liquid::C::Expression.strict_parse('obj.prop')
+    assert_equal 'some value', context.evaluate(expr)
+
+    expr = Liquid::C::Expression.strict_parse('obj["prop"]')
+    assert_equal 'some value', context.evaluate(expr)
+  end
```
macournoyer, line 58: Worth adding a test case for missing key lookup? `obj.missing`

ext/liquid_c/vm.c
```diff
@@ -284,10 +341,37 @@ static VALUE vm_render_until_error(VALUE uncast_args)
...
+            case OP_WRITE_NODE:
+                rb_funcall(cLiquidBlockBody, id_render_node, 3, args->context, output, (VALUE)*const_ptr++);
+                if (RARRAY_LEN(vm->interrupts)) {
+                    return false;
+                }
+                increment_write_score(args);
+                break;
+            case OP_RENDER_VARIABLE_RESCUE:
+                args->node_line_number = (unsigned int)*const_ptr++;
+                args->ip = ip;
```
macournoyer, line 364: `ip` and `const_ptr` are saved because next instruction is assumed to be calling a Ruby function?

macournoyer (approved):
Looking good! Just some questions.

ext/liquid_c/vm.c
dylanahsmith: This PR just moves this case to keep the rendering instruction handling together, since they shouldn't be used in the Liquid::C::Expression code.

This was added in the last PR (https://github.com/Shopify/liquid-c/pull/59) and described as follows

> Another design choice I made for this PR is to do error handling for the whole block body render, rather than for each variable render. This way we can reduce the state saving cost of `rb_rescue` from liquid code that doesn't encounter variable render errors. In order to recover from the exception, we just restore the stack size to what it was at the start of the block body render and iterate the instructions to jump just past the variable write.

In order to iterate the instructions to the end of the variable render on an exception, we need to save the instruction position at the start of variable rendering.

I'll add some comments to the code to make this clearer.

dylanahsmith: I'll also move code around in the last PR so that the diff is easier to read in this PR.

ext/liquid_c/expression.c
```diff
@@ -0,0 +1,100 @@
+#include "liquid.h"
+#include "vm_assembler.h"
+#include "parser.h"
+#include "vm.h"
+
```
peterzhu2118, line 5:
```suggestion
#include "expression.h"

```

ext/liquid_c/expression.c
```diff
@@ -0,0 +1,100 @@
...
+VALUE cLiquidCExpression;
+
+typedef struct expression {
+    vm_assembler_t code;
+} expression_t;
+
```
peterzhu2118, line 11: If we include the header, this struct should be defined.

```suggestion
```

ext/liquid_c/vm_assembler.c
```diff
@@ -25,12 +26,23 @@ void vm_assembler_gc_mark(vm_assembler_t *code)
         switch (*ip++) {
```
peterzhu2118, line 26: Can we mark the ops that need to be marked and call `liquid_vm_next_instruction` in `vm.c` to move to the next instruction?

ext/liquid_c/expression.c
dylanahsmith: Oops, I didn't mean to double include the header.  I'll fix that.

peterzhu2118 (approved):

ext/liquid_c/variable_lookup.c
```diff
@@ -2,63 +2,34 @@
...
-            Check_Type(key, T_STRING);
-            ID intern_key = rb_intern(RSTRING_PTR(key));
-            if (rb_respond_to(object, intern_key)) {
-                next_object = rb_funcall(object, rb_intern(RSTRING_PTR(key)), 0);
-                object = value_to_liquid_and_set_context(next_object, context);
-                continue;
-            }
+    if (is_command) {
+        Check_Type(key, T_STRING);
+        ID intern_key = rb_intern(RSTRING_PTR(key));
```
pushrax, line 19: Yeah it seems like we could avoid the conversion through RString here as mentioned in another comment. Definitely not a blocker, tiny optimization.

ext/liquid_c/variable_lookup.c
```diff
@@ -2,63 +2,34 @@
...
+        if (rb_respond_to(object, intern_key)) {
+            VALUE next_object = rb_funcall(object, rb_intern(RSTRING_PTR(key)), 0);
```
pushrax, line 21: Duplicate `rb_intern` here (was in the original code, probably my fault)

ext/liquid_c/vm_assembler.c
pushrax: > get rid of all non-ruby constants from the constants buffer

Where do you plan to put them? It looks like probably interleaved into the code buffer given the implementation of PUSH_INT*, which sounds good to me overall. It does make it more difficult to include medium size non-ruby constants inline. I mentioned in another comment that perhaps we could skip the conversion through RString in some cases, but I also wouldn't want to have strings in the code buffer. (in the OP_LOOKUP_COMMAND we can just use an enum, but I can imagine potentially wanting to skip RString for some drop methods and filter names).

test/unit/expression_test.rb
```diff
@@ -0,0 +1,109 @@
...
+    context = Liquid::Context.new({"x" => 123})
+    expr = Liquid::C::Expression.strict_parse('x')
+
+    assert_instance_of(Liquid::C::Expression, expr)
+    assert_equal 123, context.evaluate(expr)
+  end
+
+  def test_find_dynamic_variable
+    context = Liquid::Context.new({"x" => "y", "y" => 42})
+    expr = Liquid::C::Expression.strict_parse('[x]')
```
pushrax, line 46: I thought `[x]` on its own was supposed to be a syntax error, and that it required an identifier in front. It looks like this will make `{{ [x] }}` possible, which is currently disallowed

```rb
Liquid::Template.parse("{{ [x] }}", "x" => "y", "y" => 42, error_mode: :strict).render
...
# => Liquid::SyntaxError (Liquid syntax error: [:open_square, "["] is not a valid expression in "{{ [x] }}")
```

I'm not necessarily opposed to making it allowed, but we should get consensus from `@Shopify/guardians-of-the-liquid` and modify this in Ruby too.

pushrax (commented):
Very cool stuff!

ext/liquid_c/parser.c
```diff
@@ -67,99 +67,194 @@ static VALUE parse_number(parser_t *p)
...
+    if (RB_FIXNUM_P(num)) {
+        int x = FIX2INT(num);
+        if (x >= INT8_MIN && x <= INT8_MAX) {
+            vm_assembler_add_push_int8(code, x);
+            return;
+        } else if (x >= INT16_MIN && x <= INT16_MAX) {
+            vm_assembler_add_push_int16(code, x);
+            return;
+        }
+    }
```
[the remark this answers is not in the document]
dylanahsmith: I wanted to keep the low-level function that adds the OP_PUSH_CONST instruction.  However, `vm_assembler_add_push_literal` does selectively add an instruction based on the type, so can do this.  I'll push a commit that extracts this to `vm_assembler_add_push_fixnum` and uses it in `vm_assembler_add_push_literal`

However, I could cherry-pick an extraction of this into `vm_assembler_add_push_fixnum` which I have in a follow-up branch, which also uses this in `vm_assembler_add_push_literal`

ext/liquid_c/parser.c
```diff
@@ -169,46 +264,16 @@ VALUE parse_expression(parser_t *p)
...
-    if (p.cur.type == TOKEN_EOS)
-        return Qnil;
-
-    VALUE expr = parse_expression(&p);
-
-    if (p.cur.type != TOKEN_EOS)
-        rb_enc_raise(utf8_encoding, cLiquidSyntaxError, "[:%s] is not a valid expression", symbol_names[p.cur.type]);
-
-    return expr;
+    vm_assembler_add_push_nil(code);
```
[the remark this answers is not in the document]
dylanahsmith: The above `rb_enc_raise` calls are `NO_RETURN`, so I guess this is just dead code.  I'll remove this to avoid confusion.

ext/liquid_c/variable_lookup.c
dylanahsmith: We could optimize commands for this case, but we would still need to pass a string as the key to `rb_funcall(object, id_aref, 1, key)` if the object is a hash.

ext/liquid_c/vm_assembler.c
dylanahsmith: Yeah, I wanted to add them to the instructions buffer if they can be used in their serialized form.

I was planning on using an immediate string argument for `OP_WRITE_RAW`, since it seems rare that we would re-use the string and we won't want the deserialization overhead.

For filter, variable and key names, it is actually quite likely that we will re-use the object, so I think a table lookup would make sense.

Longer term, I'm more interested in what is most efficient, where we can re-use code to make things easier.

test/unit/expression_test.rb
dylanahsmith: Oh, I was looking at what was possible with `Liquid::Expression.parse("[x]")`.  I didn't realize we considered a dynamic variable lookup to be a strict parse error.  I change it back to being a strict parse error, since I didn't intend to change it here.

pushrax: I’ll also note that if you try to parse and render it in lax mode, it returns an empty string.

dylanahsmith:
```
~/src/liquid(master)$ bundle exec irb -Ilib -rliquid
irb> Liquid::Template.parse('{{ [x] }}').render({ 'x' => 'y', 'y' => 42 })
=> "42"
irb> context = Liquid::Context.new({ 'x' => 'y', 'y' => 42 })
irb> expression = Liquid::Expression.parse('[x]')
irb> context.evaluate(expression)
=> 42
```

dylanahsmith: Oh, it was returning an empty string for you since you were passing the variable values as an argument to parse, when they are an argument to render

pushrax: Ah, obviously 🤦

dylanahsmith: Actually, it looks like this has been the behaviour strict parse behaviour for variables in liquid-c for a long time now.  I think since variable parsing was added to liquid-c (see how `TOKEN_OPEN_SQUARE` is allowed at the start of `parse_variable` in the PR that introduced the function https://github.com/Shopify/liquid-c/pull/13/files#diff-307861feddcb5da5f0af73741d9fdea2R93)

So I'm hesitant to change it.  Especially since it could be a breaking change for strictly parsed code.  I'm also unsure if it was really the intention of the strict parser to remove any features, so we could consider whether this should be considered a strict parser bug in the liquid gem.

pushrax: I agree, in the end it seems we should add support for this to the Ruby strict parser.

pushrax (approved):

ext/liquid_c/variable.c
```diff
@@ -50,37 +35,67 @@ static VALUE try_variable_strict_parse(VALUE uncast_args)
...
             arg_count++;
-            if (RHASH_SIZE(keyword_args) > 255) {
+            if (keyword_arg_count > 255)
                 rb_enc_raise(utf8_encoding, cLiquidSyntaxError, "Too many filter keyword arguments");
-            }
-            if (const_keyword_args) {
-                vm_assembler_add_push_const(code, keyword_args);
+
+            if (const_keyword_args != Qnil) {
+                rb_hash_freeze(const_keyword_args);
```
dylanahsmith, line 89: I think we are going to have to always freeze or always dup the hash.  Otherwise, it is too easy to test a filter using dynamic keyword arguments and then get errors in production when a frozen hash is passed in for constant keyword arguments.

dylanahsmith: I've opened https://github.com/Shopify/liquid-c/pull/88 to remove the constant keyword arguments optimization on master.  We can always add it back later.

PR #64, opened 2020-10-01
Implement Liquid::ResourceLimits

peterzhu2118:
Implements `Liquid::ResourceLimits` in C. It is roughly a translation of the Ruby code into C. I was not able to measure a significant speedup on the benchmarks.

### DIFF
--- a/ext/liquid_c/resource_limits.c
+++ b/ext/liquid_c/resource_limits.c
@@ -2,7 +2,6 @@
 #include "resource_limits.h"
 
 VALUE cLiquidResourceLimits;
-static VALUE id_bytesize;
 
 static void resource_limits_free(void *ptr)
 {
@@ -179,8 +178,6 @@ static VALUE resource_limits_increment_assign_score_method(VALUE self, VALUE amo
 
 void resource_limits_increment_write_score(resource_limits_t *resource_limits, VALUE output)
 {
-    Check_Type(output, T_STRING);
-
     long captured = RSTRING_LEN(output);
 
     if (resource_limits->last_capture_length >= 0) {
@@ -194,6 +191,8 @@ void resource_limits_increment_write_score(resource_limits_t *resource_limits, V
 
 static VALUE resource_limits_increment_write_score_method(VALUE self, VALUE output)
 {
+    Check_Type(output, T_STRING);
+
     resource_limits_t *resource_limits;
     ResourceLimits_Get_Struct(self, resource_limits);
 
@@ -257,8 +256,6 @@ static VALUE resource_limits_reset_method(VALUE self)
 
 void init_liquid_resource_limits()
 {
-    id_bytesize = rb_intern("bytesize");
-
     cLiquidResourceLimits = rb_define_class_under(mLiquidC, "ResourceLimits", rb_cObject);
     rb_global_variable(&cLiquidResourceLimits);
 
--- a/ext/liquid_c/resource_limits.c
+++ b/ext/liquid_c/resource_limits.c
@@ -1,0 +1,283 @@
+#include "liquid.h"
+#include "resource_limits.h"
+
+VALUE cLiquidResourceLimits;
+static VALUE id_plus, id_minus, id_gt, id_bytesize;
+
+static void resource_limits_mark(void *ptr)
+{
+    resource_limits_t *resource_limits = ptr;
+    rb_gc_mark(resource_limits->render_length_limit);
+    rb_gc_mark(resource_limits->render_score_limit);
+    rb_gc_mark(resource_limits->assign_score_limit);
+    rb_gc_mark(resource_limits->last_capture_length);
+    rb_gc_mark(resource_limits->render_score);
+    rb_gc_mark(resource_limits->assign_score);
+}
+
+static void resource_limits_free(void *ptr)
+{
+    resource_limits_t *resource_limits = ptr;
+    xfree(resource_limits);
+}
+
+static size_t resource_limits_memsize(const void *ptr)
+{
+    return ptr ? sizeof(resource_limits_t) : 0;
+}
+
+const rb_data_type_t resource_limits_data_type = {
+    "liquid_resource_limits",
+    { resource_limits_mark, resource_limits_free, resource_limits_memsize },
+    NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+static VALUE resource_limits_allocate(VALUE klass)
+{
+    resource_limits_t *resource_limits;
+
+    VALUE obj = TypedData_Make_Struct(klass, resource_limits_t, &resource_limits_data_type, resource_limits);
+
+    return obj;
+}
+
+static void resource_limits_reset(resource_limits_t *resource_limit)
+{
+    resource_limit->reached_limit = 0;
+    resource_limit->last_capture_length = Qnil;
+    resource_limit->render_score = INT2FIX(0);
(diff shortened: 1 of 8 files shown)


ext/liquid_c/vm.c
```diff
@@ -17,13 +15,16 @@ void liquid_vm_render(block_body_t *body, VALUE context, VALUE output)
     Check_Type(output, T_STRING);
 
     VALUE resource_limits = rb_ivar_get(context, id_ivar_resource_limits);
+    resource_limits_t *resource_limits_struct;
+    ResourceLimits_Get_Struct(resource_limits, resource_limits_struct);
```
dylanahsmith, line 19: We shouldn't even need the `resource_limits` object after we get the struct pointer out of it, so we may as well not keep it accessible after this

```suggestion
    resource_limits_t *resource_limits;
    ResourceLimits_Get_Struct(rb_ivar_get(context, id_ivar_resource_limits), resource_limits);
```

Specifically, we want to replace the `rb_funcall(resource_limits, id_increment_write_score, 1, output);` line, since it would be on a hot path within the capture tag.

ext/liquid_c/vm.c
```diff
@@ -67,8 +68,6 @@ void init_liquid_vm()
     id_render_node = rb_intern("render_node");
     id_ivar_interrupts = rb_intern("@interrupts");
     id_ivar_resource_limits = rb_intern("@resource_limits");
-    id_ivar_render_length_limit = rb_intern("@render_length_limit");
-    id_ivar_last_capture_length = rb_intern("@last_capture_length");
     id_increment_render_score = rb_intern("increment_render_score");
     id_increment_write_score = rb_intern("increment_write_score");
     id_raise_limits_reached = rb_intern("raise_limits_reached");
```
dylanahsmith, line 73: These shouldn't be needed here either, since we can instead do C function calls now that we have a resource limits struct.

dylanahsmith (changes requested):
I think the performance improvement would mostly be significant within a capture tag at the moment.  So you could try temporarily modifying the benchmark to wrap the benchmarked templates in a capture tag.

I expect it will also help with the assign tag and to reduce the instance variable lookup overhead on the resource limits object.

peterzhu2118:
Thanks for the feedback @dylanahsmith! I changed the benchmark files to use capture tags, I think there is a slight improvement in performance, however the results are pretty noisy on my computer.

Master:
```
              parse:    129.494  (±10.0%) i/s -      1.287k in  10.046882s
             render:    135.550  (± 8.9%) i/s -      1.356k in  10.093187s
     parse & render:     62.487  (± 6.4%) i/s -    624.000  in  10.033199s
```

This branch:
```
              parse:    131.418  (± 6.8%) i/s -      1.313k in  10.041110s
             render:    149.923  (± 8.7%) i/s -      1.498k in  10.075368s
     parse & render:     65.270  (± 6.1%) i/s -    654.000  in  10.067960s
```

dylanahsmith:
Wow, that is a more significant difference in the benchmark than I expected.

ext/liquid_c/resource_limits.c
```diff
@@ -0,0 +1,283 @@
...
+    { resource_limits_mark, resource_limits_free, resource_limits_memsize },
+    NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+static VALUE resource_limits_allocate(VALUE klass)
+{
+    resource_limits_t *resource_limits;
+
+    VALUE obj = TypedData_Make_Struct(klass, resource_limits_t, &resource_limits_data_type, resource_limits);
+
```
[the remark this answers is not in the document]
dylanahsmith: I still think we should move `resource_limits_reset(resource_limits);` from the initialize method to this allocate function, which will also remove the need for having to use `ResourceLimits_Get_Struct(self, resource_limits);` there.

ext/liquid_c/vm.c
```diff
@@ -54,9 +51,9 @@ void liquid_vm_render(block_body_t *body, VALUE context, VALUE output)
                 rb_bug("invalid opcode: %u", ip[-1]);
         }
         if (RB_UNLIKELY(is_captured)) {
-            rb_funcall(resource_limits, id_increment_write_score, 1, output);
+            resource_limits_increment_write_score(resource_limits, output);
         } else if (RSTRING_LEN(output) > render_length_limit) {
-            rb_funcall(resource_limits, id_raise_limits_reached, 0);
+            resource_limits_raise_limits_reached(resource_limits);
         }
```
dylanahsmith, line 57: Semantically, this is `increment_write_score`, just written like this as an optimization.  I extracted it to an inline `increment_write_score` function in https://github.com/Shopify/liquid-c/pull/59 which we could probably move to resource_limits.h and just have a pointer to the resource_limits struct and the output string as parameters.

I don't think we need the `is_captured` or `render_length_limit` local variables in this function now, they were just here to reduce ruby function calls.

ext/liquid_c/resource_limits.c
```diff
@@ -0,0 +1,282 @@
...
+    ResourceLimits_Get_Struct(self, resource_limits);
+
+    resource_limits_increment_assign_score(resource_limits, NUM2LONG(amount));
+
+    return Qnil;
+}
+
+void resource_limits_increment_write_score(resource_limits_t *resource_limits, VALUE output)
+{
+    long captured = NUM2LONG(rb_funcall(output, id_bytesize, 0));
```
dylanahsmith, line 185: We should do the ruby object checking and conversion in the ruby method function (e.g. resource_limits_increment_write_score_method in this case) so we avoid this extra work when calling from C code.  As part of that type check, we should use `Check_Type(output, T_STRING);` so we can get the byte size using `RSTRING_LEN(output)` here.  Note that `RSTRING_LEN` does get the byte size despite its name implying it gets length.

ext/liquid_c/vm.h
```diff
@@ -3,6 +3,7 @@
 
 #include <ruby.h>
 #include "block.h"
+#include "resource_limits.h"
```
dylanahsmith, line 6: It doesn't look like this is needed for this header, so this include could be moved into `vm.c`

dylanahsmith (changes requested):

ext/liquid_c/resource_limits.c
```diff
@@ -0,0 +1,282 @@
...
+    ResourceLimits_Get_Struct(self, resource_limits);
+
+    resource_limits_increment_assign_score(resource_limits, NUM2LONG(amount));
+
+    return Qnil;
+}
+
+void resource_limits_increment_write_score(resource_limits_t *resource_limits, VALUE output)
+{
+    Check_Type(output, T_STRING);
```
dylanahsmith, line 182: In https://github.com/Shopify/liquid-c/pull/64#discussion_r498825055 I meant that this should be done in resource_limits_increment_write_score_method to avoid doing it redundantly in the VM loop.

ext/liquid_c/resource_limits.c
```diff
@@ -0,0 +1,282 @@
...
+{
+    resource_limits_t *resource_limits;
+    ResourceLimits_Get_Struct(self, resource_limits);
+    resource_limits_reset(resource_limits);
+    return Qnil;
+}
+
+void init_liquid_resource_limits()
+{
+    id_bytesize = rb_intern("bytesize");
```
dylanahsmith, line 260: This is unused, so can be removed now.

dylanahsmith (approved):
A couple of minor things, otherwise, LGTM

Issue #65, opened 2020-10-02
Encoding::CompatibilityError when using a binary encoded output


dylanahsmith:
We were getting the error "incompatible character encodings: ASCII-8BIT and UTF-8 (Encoding::CompatibilityError)" after https://github.com/Shopify/liquid-c/pull/58 was introduced when trying to render using a binary encoded output buffer.

This error can be demonstrated with the following code

```
output = String.new(capacity: 10) # defaults to encoding: Encoding::BINARY
template = Liquid::Template.parse("ü{{ unicode_char }}")
template.render!({ 'unicode_char' => 'ñ' }, output: output) # raises Encoding::CompatibilityError
```

This was happening because the OP_WRITE_RAW instruction was writing using `rb_str_cat` and a character pointer, which isn't encoding aware, so was writing the UTF-8 text to the binary string.  After that, the output buffer is no longer ascii compatible or UTF-8 compatible, so will get a compatibility error when a unicode encoded string with non-ascii compatible characters is written to the output buffer, as happens when writing a string variable.

PR #66, opened 2020-10-02
Ensure template source and output strings are UTF-8 encoded

dylanahsmith:
Fixes https://github.com/Shopify/liquid-c/issues/65

When one ruby string is appended to another, ruby will automatically do encoding conversion as long as the two strings are compatible.  This actually hides bugs in unicode character handling when testing with ascii compatible, leading to the issue this PR fixes.

Instead, I would like liquid-c to be stricter than ruby about the encoding of the input strings, so we can catch encoding errors easier during testing.  This PR simply checks to make sure the source and output encodings are exactly UTF-8 encoded and raises if they aren't.

### DIFF
--- a/test/unit/block_test.rb
+++ b/test/unit/block_test.rb
@@ -8,30 +8,4 @@ class BlockTest < MiniTest::Test
     template = Liquid::Template.parse("{{ -}} foo {{- }}")
     assert_equal 3, template.root.nodelist.size
   end
-
-  def test_pre_trim
-    template = Liquid::Template.parse("\n{%- raw %}{% endraw %}")
-    assert_equal "", template.render
-
-    template = Liquid::Template.parse("\n{%- if true %}{% endif %}")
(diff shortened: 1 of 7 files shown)


test/unit/block_test.rb
```diff
@@ -9,29 +9,18 @@ def test_no_allocation_of_trimmed_strings
     assert_equal 3, template.root.nodelist.size
   end
 
-  def test_pre_trim
```
dylanahsmith, line 12: This test and test_bug_compatible_pre_trim were moved to the shared liquid integration tests (https://github.com/Shopify/liquid/pull/1291/files#diff-abd62f5ce9d5df8c4bce6644f57a9b7a) so I removed them here.

pushrax (approved):
Approach sounds good to me.

peterzhu2118 (approved):

PR #68, opened 2020-10-05
Fix infinite loop in the buffer when capacity is set to 0

peterzhu2118:
If we try to allocate a buffer with capacity set to 0, resizing it will cause an infinite loop because `new_capacity *= 2;` will always be `0`.


dylanahsmith:
I actually had a commit to fix this (https://github.com/Shopify/liquid-c/commit/80d8587bc3e9cb855eca8809160bcf0ae416c1e0) in https://github.com/Shopify/liquid-c/pull/59

I could extract that so it could be reviewed and merged in isolation.

peterzhu2118:
@dylanahsmith That would be great!

peterzhu2118:
Closing because #70 is merged.

PR #70, opened 2020-10-05
Refactor c_buffer to make it suitable to re-use for a stack

dylanahsmith:
Extracted from https://github.com/Shopify/liquid-c/pull/59

> * I changed the c_buffer struct to store pointers for the end of the data and capacity, instead of storing the sizes and capacity directly. This is because we are mostly getting and updating the data end pointer, which is especially true in order to use this for the VM stack where we reserve space for the block and don't have to check the capacity for each write/push.
> * I've also added support to c_buffer to be initialized with size 0 (without an allocation) which is the one case where we can't just double the capacity until we have the requested extra capacity.

### DIFF
--- a/ext/liquid_c/vm_assembler.h
+++ b/ext/liquid_c/vm_assembler.h
@@ -22,7 +22,7 @@ void vm_assembler_add_write_node(vm_assembler_t *code, VALUE node);
 
 static inline size_t vm_assembler_alloc_memsize(const vm_assembler_t *code)
 {
-    return code->instructions.capacity + code->constants.capacity;
+    return c_buffer_capacity(&code->instructions) + c_buffer_capacity(&code->constants);
 }
 
 static inline void vm_assembler_write_opcode(vm_assembler_t *code, enum opcode op)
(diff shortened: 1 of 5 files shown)


ext/liquid_c/vm_assembler.h
```diff
@@ -42,7 +42,7 @@ static inline void vm_assembler_add_leave(vm_assembler_t *code)
 
 static inline void vm_assembler_remove_leave(vm_assembler_t *code)
 {
-    code->instructions.size -= 1;
+    code->instructions.data_end--;
```
peterzhu2118, line 45: Instead of just decrementing the counter, should we have a function that decrements the size of the buffer which could also realloc and free some memory if we remove beyond a certain threshold?

peterzhu2118 (approved):
One minor comment, otherwise looks good!

ext/liquid_c/vm_assembler.h
dylanahsmith: This would be a bad place to shrink the buffer, since the leave instruction is being removed in order to extend the buffer after which the leave instruction will be added again.

PR #71, opened 2020-10-06
Implement Liquid::Raw#parse

peterzhu2118:
Implement `Liquid::Raw#parse`.

The function `match_full_token_possibly_invalid` emulates matching the `FullTokenPossiblyInvalid = /\A(.*)#{TagStart}\s*(\w+)\s*(.*)?#{TagEnd}\z/om` regex.

# Benchmark

For the benchmark, I wrapped every `theme.liquid` in a `raw` tag and ran the benchmark.

Master:

```
              parse:    175.223  (± 9.7%) i/s -      1.746k in  10.078290s
             render:    250.375  (± 8.4%) i/s -      2.496k in  10.042677s
     parse & render:     96.516  (± 7.3%) i/s -    960.000  in  10.009864s
```

This branch:

```
              parse:    280.836  (± 9.3%) i/s -      2.793k in  10.040895s
             render:    236.161  (±12.3%) i/s -      2.340k in  10.076232s
     parse & render:    120.075  (± 9.2%) i/s -      1.199k in  10.070105s
```

In this (very) synthetic benchmark, we see a 60% increase in performance during parsing.

### DIFF
--- a/ext/liquid_c/raw.c
+++ b/ext/liquid_c/raw.c
@@ -1,0 +1,114 @@
+#include "liquid.h"
+#include "raw.h"
+#include "stringutil.h"
+#include "tokenizer.h"
+
+VALUE cLiquidRaw;
+static VALUE id_parse_context, id_locale, id_translate, id_block_name, id_block_delimiter, id_ivar_body;
+
+__attribute__((noreturn)) static void raise_translated_syntax_error(VALUE self, const char *key, VALUE args)
+{
+    VALUE key_str = rb_str_new_cstr(key);
+    VALUE msg = rb_funcall(rb_funcall(rb_funcall(self, id_parse_context, 0), id_locale, 0), id_translate, 2, key_str, args);
+    rb_raise(cLiquidSyntaxError, "%.*s", (int)RSTRING_LEN(msg), RSTRING_PTR(msg));
+}
+
+struct full_token_possibly_invalid_t {
+    const char *body_start;
+    long body_len;
+    const char *delimiter_start;
+    long delimiter_len;
+};
+
+static bool match_full_token_possibly_invalid(token_t *token, struct full_token_possibly_invalid_t *match)
+{
+    const char *str = token->str_full;
+    long len = token->len_full;
+
+    match->body_start = str;
+    match->body_len = 0;
+    match->delimiter_start = NULL;
+    match->delimiter_len = 0;
+
+    if (len < 5) return false; // Must be at least 5 characters: \{%\w%\}
+    if (str[len - 1] != '}' || str[len - 2] != '%') return false;
+
+    const char *curr_delimiter_start;
+    long curr_delimiter_len = 0;
+
+    for (long i = len - 3; i >= 0; i--) {
+        char c = str[i];
+
+        if (is_word(c)) {
+            curr_delimiter_start = str + i;
+            curr_delimiter_len++;
+        } else {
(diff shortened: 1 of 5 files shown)


ext/liquid_c/raw.c
```diff
@@ -0,0 +1,114 @@
...
+    long block_delimiter_len = RSTRING_LEN(block_delimiter);
+
+    while (true) {
+        tokenizer_next(tokenizer, &token);
+
+        if (!token.type) break;
+
+        if (match_full_token_possibly_invalid(&token, &match)
+                && match.delimiter_len == block_delimiter_len
+                && strncmp(match.delimiter_start, block_delimiter_str, block_delimiter_len) == 0) {
```
dylanahsmith, line 86: We could use memcmp, since we don't really want to special case null characters, which would also slow down the comparison

```suggestion
                && memcmp(match.delimiter_start, block_delimiter_str, block_delimiter_len) == 0) {
```

ext/liquid_c/raw.c
```diff
@@ -0,0 +1,114 @@
...
+            rb_ivar_set(self, id_ivar_body, body_str);
+            return Qnil;
+        }
+
+        c_buffer_write(&body, (char *)token.str_full, token.len_full);
+    }
+
+    VALUE hash = rb_hash_new();
+    rb_hash_aset(hash, id_block_name, rb_funcall(self, id_block_name, 0));
+    raise_translated_syntax_error(self, "errors.syntax.tag_never_closed", hash);
```
dylanahsmith, line 98: An alternative would be to call a ruby method to raise the exception.  We could even refactor the ruby gem to expose that method so we don't have to define it in lib/liquid/c.rb.  That is what we have done in other places where a translated error is raised.

Besides, the ruby calls in `raise_translated_syntax_error` are going to be slow than the same ones done from ruby code, since the ruby VM uses an inline cache for those calls that it doesn't expose to C extensions.

ext/liquid_c/raw.c
```diff
@@ -0,0 +1,114 @@
...
+{
+    id_parse_context = rb_intern("parse_context");
+    id_locale = rb_intern("locale");
+    id_translate = rb_intern("t");
+    id_block_name = rb_intern("block_name");
+    id_block_delimiter = rb_intern("block_delimiter");
+    id_ivar_body = rb_intern("@body");
+
+    cLiquidRaw = rb_const_get(mLiquid, rb_intern("Raw"));
+    rb_global_variable(&cLiquidRaw);
```
dylanahsmith, line 111: `cLiquidRaw` isn't needed outside of this function, so we could just make it local

```suggestion
    VALUE cLiquidRaw = rb_const_get(mLiquid, rb_intern("Raw"));
```

ext/liquid_c/raw.h
```diff
@@ -0,0 +1,18 @@
...
+
+typedef struct raw {
+    VALUE tag_name;
+    VALUE parse_context;
+    c_buffer_t body;
+} raw_t;
+
+extern VALUE cLiquidRaw;
+extern const rb_data_type_t raw_data_type;
+#define Raw_Get_Struct(obj, sval) TypedData_Get_Struct(obj, raw_t, &raw_data_type, sval)
```
dylanahsmith, line 14: We don't need any of this in the header file.

dylanahsmith (changes requested):

dylanahsmith (approved):

Issue #77, opened 2020-10-08
Add support for serializing liquid templates along with their VM code


dylanahsmith:
Once https://github.com/Shopify/liquid-c/pull/59 and https://github.com/Shopify/liquid-c/pull/60 are merged, the liquid benchmarks show that 60% of the time to parse and render a template happens during parsing.  We could save a lot of that time by persisting (e.g. to a cache initially) the parsed template so that it can be re-used for further renders.

I think we should have the persisted format be fairly stable, such that it can be used across deploys of the application.  As such, I don't think we should just use `Marshal` on the liquid gem's parse tree, since that can easily lead to breaking changes from code written without awareness of liquid-c.  Instead, we should re-parse tag nodes on deserialization if that don't support liquid VM compilation.  For code that is compiled into liquid VM code, we should include two versions in the serialized VM code, one for the liquid-c and one for the application, so either can bump these versions to avoid rendering incompatible VM code.

So the serialized block bodies will consist of the following sections:
1. liquid VM instructions that don't require deserialization for execution
2. An array of marshal dumped ruby constants and will require marshal loading on deserialization
3. Uncompiled tags to parse on deserialization, including the type of tag (e.g. regular, block or raw) and references to any block bodies and "unknown tags" (i.e. handled from Liquid::Block#unknown_tag, such as `elsif` for Liquid::If)

This will require combining the VM assembler buffers for these block bodies together and removing the assumption that each block body will own its own assembler buffers.  Instead, the block bodies should instead reference offsets into the shared buffer for execution.

To keep the ruby constant array simple, we will want to compile the non-ruby constant arguments into the VM instructions as immediate arguments.  Currently this isn't the case for OP_RENDER_VARIABLE_RESCUE and OP_WRITE_RAW.  Note that doing this for `OP_WRITE_RAW` will add copying overhead that won't be beneficial without using it for serialization.

Another possible application version compatibility concern is the Liquid::ParseContext, which is an argument for parsing, which can end up being saved and used for rendering (e.g. for parsing included templates).  To check for compatibility, we may want to require the application to pass in the same parse options for both parsing and deserializing to consider the serialized template to be compatible with the current application version.  We may also want to check the parse context after parsing to make sure the application doesn't depend on mutations of the parse context.

tobi:
[...]

PR #79, opened 2020-10-09
Implement OP_RENDER_VARIABLE_RESCUE on the instructions buffer

peterzhu2118:
This implements storing the line number of OP_RENDER_VARIABLE_RESCUE instruction on the instructions buffer rather than the constants buffer. We only need to support line number of up to 24 bytes so it's stored as 3 x 8 byte values. 

See #77.

### DIFF
--- a/ext/liquid_c/vm.c
+++ b/ext/liquid_c/vm.c
@@ -194,7 +194,7 @@ typedef struct vm_render_until_error_args {
     vm_t *vm;
     const uint8_t *ip; // use for initial address and to save an address for rescuing
     const size_t *const_ptr;
-    unsigned int node_line_number;
+    uint8_t node_line_number[3];
     VALUE context;
     VALUE output;
 } vm_render_until_error_args_t;
@@ -273,7 +273,7 @@ static VALUE vm_render_until_error(VALUE uncast_args)
                 break;
             case OP_RENDER_VARIABLE_RESCUE:
                 // Save state used by vm_render_rescue to rescue from a variable rendering exception
-                args->node_line_number = (ip[0] << 16) | (ip[1] << 8) | ip[2];
+                memcpy(&args->node_line_number, ip, sizeof(args->node_line_number));
                 // vm_render_rescue will iterate from this instruction to the instruction
                 // following OP_POP_WRITE_VARIABLE to resume rendering from
                 ip += 3;
@@ -340,6 +340,11 @@ typedef struct vm_render_rescue_args {
     size_t old_stack_byte_size;
 } vm_render_rescue_args_t;
 
+static inline unsigned int decode_node_line_number(uint8_t *node_line_number)
+{
+    return (node_line_number[0] << 16) | (node_line_number[1] << 8) | node_line_number[2];
+}
+
 // Actually returns a bool resume_rendering value
 static VALUE vm_render_rescue(VALUE uncast_args, VALUE exception)
 {
@@ -370,7 +375,8 @@ static VALUE vm_render_rescue(VALUE uncast_args, VALUE exception)
         vm->invoking_filter = false;
     }
 
-    VALUE line_number = render_args->node_line_number == 0 ? Qnil : UINT2NUM(render_args->node_line_number);
+    unsigned int node_line_number = decode_node_line_number(render_args->node_line_number);
+    VALUE line_number = node_line_number == 0 ? Qnil : UINT2NUM(node_line_number);
 
     rb_funcall(cLiquidBlockBody, rb_intern("c_rescue_render_node"), 5,
         render_args->context, render_args->output, line_number, exception, blank_tag);
--- a/ext/liquid_c/vm.c
+++ b/ext/liquid_c/vm.c
@@ -194,7 +194,7 @@ typedef struct vm_render_until_error_args {
     vm_t *vm;
     const uint8_t *ip; // use for initial address and to save an address for rescuing
     const size_t *const_ptr;
-    uint8_t node_line_number[3];
+    const uint8_t *node_line_number;
     VALUE context;
     VALUE output;
 } vm_render_until_error_args_t;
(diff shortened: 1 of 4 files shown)


ext/liquid_c/vm_assembler.h
```diff
@@ -101,8 +101,10 @@ static inline void vm_assembler_add_filter(vm_assembler_t *code, VALUE filter_na
 
 static inline void vm_assembler_add_render_variable_rescue(vm_assembler_t *code, size_t node_line_number)
 {
-    c_buffer_write(&code->constants, &node_line_number, sizeof(size_t));
-    vm_assembler_write_opcode(code, OP_RENDER_VARIABLE_RESCUE);
+    assert(node_line_number < (1 << 24));
```
dylanahsmith, line 104: We should raise a ruby exception if something isn't supported.

ext/liquid_c/vm.c
```diff
@@ -273,9 +273,10 @@ static VALUE vm_render_until_error(VALUE uncast_args)
                 break;
             case OP_RENDER_VARIABLE_RESCUE:
                 // Save state used by vm_render_rescue to rescue from a variable rendering exception
-                args->node_line_number = (unsigned int)*const_ptr++;
+                args->node_line_number = (ip[0] << 16) | (ip[1] << 8) | ip[2];
```
dylanahsmith, line 276: We don't really need to decode the number immediately.  The number is only needed when rescuing from an exception.  We could instead store a pointer to the line number bytes here, then decode them in `vm_render_rescue`

ext/liquid_c/vm_assembler.h
```diff
@@ -101,8 +102,12 @@ static inline void vm_assembler_add_filter(vm_assembler_t *code, VALUE filter_na
 
 static inline void vm_assembler_add_render_variable_rescue(vm_assembler_t *code, size_t node_line_number)
 {
-    c_buffer_write(&code->constants, &node_line_number, sizeof(size_t));
-    vm_assembler_write_opcode(code, OP_RENDER_VARIABLE_RESCUE);
+    if (RB_UNLIKELY(node_line_number >= (1 << 24))) {
+        rb_enc_raise(utf8_encoding, rb_eRuntimeError, "Line number %ld is too large", node_line_number);
```
dylanahsmith, line 106: I think `ArgumentError` (rb_eArgError) would be more appropriate existing exception class for this.

"Line number too large 16777217" might be a confusing error message.  It seems like an error message that makes sense to the caller of `vm_assembler_add_render_variable_rescue`, but this is a ruby exception being presented to the caller of `Liquid::Template.parse`.  It might be clearer if the error message was related to the input to that.

Should we instead check the source code length?  That way we could easily check it upfront instead of throughout parsing.  We could also state more clearly that only a template source of up to 16 MiB is supported.  That would also allow us to use that assumption elsewhere.

ext/liquid_c/vm.c
```diff
@@ -273,9 +273,10 @@ static VALUE vm_render_until_error(VALUE uncast_args)
                 break;
             case OP_RENDER_VARIABLE_RESCUE:
                 // Save state used by vm_render_rescue to rescue from a variable rendering exception
-                args->node_line_number = (unsigned int)*const_ptr++;
+                memcpy(&args->node_line_number, ip, sizeof(args->node_line_number));
```
macournoyer, line 276: Nice! Had to check disassembly. I would have assumed `args->node_line_number[0] = ...` would have been better, but it's not.

But it is slower than the previous code.

macournoyer (commented):
I benchmarked against master, and it seems this is a bit slower (~3000 i/s less on rendering).

@dylanahsmith I know you want to keep the ruby constant array simple, but do you think this is something to worry about ^^

Shouldn't execution/rendering time be prioritized over deserialization time? I guess it depends on the template, so I'm not sure on this...

ext/liquid_c/vm.c
dylanahsmith: We could try just storing a pointer to the bytes, as I suggested in https://github.com/Shopify/liquid-c/pull/79#discussion_r502504712 .  That seems like it might require less instructions than copying 3 individual bytes.

dylanahsmith:
> Shouldn't execution/rendering time be prioritized over deserialization time? I guess it depends on the template, so I'm not sure on this...

I would like to prioritize the combined time to deserialize and render the template.  Yes, it will depend on the template, so there can definitely end up being trade-offs in some cases.

ext/liquid_c/vm.c
macournoyer: Checked how other VMs do it. I think we should have two instructions: `OP_RENDER_VARIABLE_RESCUE` & `OP_RENDER_VARIABLE_RESCUE_W` (wide version).

The wide version does the decode over the 3 bytes, directly in the `case`, the non-wide one only use one byte (and no bitshifting).

Similar to the `LOADK` & `LOADKX` (wide version) instructions in Lua.  And inspired by Java too, that has many wide (`*_w`) version of instructions: https://en.wikipedia.org/wiki/Java_bytecode_instruction_listings. Also, Java has a whole other category of wide instructions, see the `wide` at the bottom, very interesting.

dylanahsmith: I still think doing any decoding is a waste, since this is just a line number that is only needed on the error path.  I believe MRI ruby and C++ take the approach of keeping the exception handling out of the normal execution path and then just looking up data to handle the rescue path based on the instruction position when an exception is being handled.

The use of the rescue instruction was for simplicity before we need more general rescue handling.

peterzhu2118: I did a quick implementation of `OP_RENDER_VARIABLE_RESCUE` and `OP_RENDER_VARIABLE_RESCUE_W` instruction to benchmark. It didn't seem to perform any faster on the benchmark.

peterzhu2118: I pushed it onto the `pz-render-var-rescue-wide` branch if anyone wants to play around with it.

macournoyer: Ah but I see you removed the `memcpy`. It does indeed remove the need for the wide instruction. That's a brilliant idea 👏

dylanahsmith (approved):

PR #81, opened 2020-10-09
Require source code to be at most 16MiB

peterzhu2118:
Raise when source is larger than 16MiB in `Liquid::Template#parse`. This will let us use this assumption in liquid-c.

See #79.

### DIFF
--- a/ext/liquid_c/tokenizer.c
+++ b/ext/liquid_c/tokenizer.c
@@ -47,6 +47,12 @@ static VALUE tokenizer_initialize_method(VALUE self, VALUE source, VALUE start_l
     Check_Type(source, T_STRING);
     check_utf8_encoding(source, "source");
 
+#define MAX_SOURCE_CODE_BYTES ((1 << 24) - 1)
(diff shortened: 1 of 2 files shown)


lib/liquid/c.rb
```diff
@@ -211,6 +211,19 @@ def c_find_variable_kwarg(key, raise_on_not_found: true)
   end
 end
 
+Liquid::Template.const_set("MAX_SOURCE_CODE_BYTES", (2**24) - 1)
+Liquid::Template.class_eval do
+  alias_method :ruby_parse, :parse
+
+  def c_parse(source, options = {})
+    if source.bytesize > Liquid::Template::MAX_SOURCE_CODE_BYTES
```
dylanahsmith, line 219: I think it would be simpler do this is on an existing parsing code path, rather than introducing another monkey patch.  Also, we generally do the safety checks for C code in the C code itself, which avoids the ability for ruby code to trigger undefined behaviour.  So I would do it in `tokenizer_initialize_method` after the existing `source` checks.

dylanahsmith (approved):

PR #85, opened 2020-10-13
Provide a clearer error message for non-string convertible variable results

dylanahsmith:
## Problem

When writing the variable result, the object might first need to be converted to a string.  If that results in an exception, then we get a hard to debug exception that starts with

```
bundler/gems/liquid-c-fe47cb400e86/lib/liquid/c.rb:9:in `render_to_output_buffer': no implicit conversion of nil into String (TypeError)
    from bundler/gems/liquid-c-fe47cb400e86/lib/liquid/c.rb:9:in `render'
```

which shows that it comes from liquid VM rendering, but it doesn't provide enough context on what it was trying to do.  That exception comes from the `StringValue` macro call ruby does in `rb_str_append`.

## Solution

In this situation, the problem is actually with the return value of the `to_s` method, which was being used with `rb_str_append`.  So we want to the exception to be clear about the source of this exception, including specifying the type of the object `to_s` is called on so that we know what method needs fixing.

### DIFF
--- a/ext/liquid_c/vm.c
+++ b/ext/liquid_c/vm.c
@@ -9,7 +9,6 @@ ID id_render_node;
 ID id_ivar_interrupts;
 ID id_ivar_resource_limits;
 ID id_to_s;
-ID id_to_str;
 ID id_vm;
 ID id_strainer;
 ID id_filter_methods_hash;
@@ -114,28 +113,15 @@ static void write_fixnum(VALUE output, VALUE fixnum)
     snprintf(RSTRING_PTR(output) + old_size, write_length + 1, "%lld", number);
 }
 
-static VALUE obj_check_to_str(VALUE obj)
-{
-    VALUE str = rb_check_funcall(obj, id_to_str, 0, NULL);
-    if (RB_UNLIKELY(str != Qundef && !RB_TYPE_P(str, T_STRING))) {
-        rb_raise(rb_eTypeError, "%"PRIsVALUE"#to_str returned a non-String value of type %"PRIsVALUE,
-                rb_obj_class(obj), rb_obj_class(str));
-    }
-    return str;
-}
-
 static VALUE obj_to_s(VALUE obj)
 {
     VALUE str = rb_funcall(obj, id_to_s, 0);
-    if (RB_TYPE_P(str, T_STRING))
+
+    if (RB_LIKELY(RB_TYPE_P(str, T_STRING)))
         return str;
 
-    VALUE to_str_result = obj_check_to_str(str);
-    if (RB_UNLIKELY(to_str_result == Qundef)) {
-        rb_raise(rb_eTypeError, "%"PRIsVALUE"#to_s returned a non-String convertible value of type %"PRIsVALUE,
-                rb_obj_class(obj), rb_obj_class(str));
-    }
-    return to_str_result;
+    rb_raise(rb_eTypeError, "%"PRIsVALUE"#to_s returned a non-String convertible value of type %"PRIsVALUE,
+            rb_obj_class(obj), rb_obj_class(str));
 }
 
 static void write_obj(VALUE output, VALUE obj)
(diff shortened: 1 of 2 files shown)


ext/liquid_c/vm.c
```diff
@@ -113,9 +114,36 @@ static void write_fixnum(VALUE output, VALUE fixnum)
...
+    return str;
+}
+
+static VALUE obj_to_s(VALUE obj)
+{
+    VALUE str = rb_funcall(obj, id_to_s, 0);
+    if (RB_TYPE_P(str, T_STRING))
+        return str;
+
+    VALUE to_str_result = obj_check_to_str(str);
```
peterzhu2118, line 133: I'm curious as to why we try to call `to_str` on the return value of `to_s` if it does not return a string? Isn't it a bug in the caller if the object `to_s` does not return a string?

dylanahsmith: This preserves the behaviour that ruby liquid has with `output << obj.to_s` where `to_s` is called explicitly, and `String#<<` calls `to_str` on the result.

An example I know about where this happens is `ActiveSupport::SafeBuffer`.  It's `#to_s` method returns itself, which seems reasonable given that `ActiveSupport::SafeBuffer` inherits from `String`.  However, it's `#to_str` (inherited from String#to_str) creates a String instance with its content duplicated.

ext/liquid_c/vm.c
```diff
@@ -138,10 +166,6 @@ static void write_obj(VALUE output, VALUE obj)
             break;
         case T_NIL:
             break;
-        default:
-            obj = rb_funcall(obj, id_to_s, 0);
-            rb_str_append(output, obj);
```
paracycle, line 143: Who is doing this append in the new code path?

ext/liquid_c/vm.c
peterzhu2118: I think `String#<<` only calls `to_str` if `to_s` does not return a string-like value. For example:

This class `IsString` is string-like, so it does not call `to_str`:
```
irb(main):001:1* class IsString < String
irb(main):002:2*   def to_s
irb(main):003:2*     self
irb(main):004:1*   end
irb(main):005:2*   def to_str
irb(main):006:2*     "bar"
irb(main):007:1*   end
irb(main):008:0> end
=> :to_str
irb(main):009:0> f = IsString.new("foo")
irb(main):010:0> s = ''
irb(main):011:0> s << f
=> "foo"
```

But this class `IsNotString` is not string-like, so it does call `to_str`:
```
irb(main):001:1* class IsNotString
irb(main):002:2*   def to_s
irb(main):003:2*     self
irb(main):004:1*   end
irb(main):005:2*   def to_str
irb(main):006:2*     "bar"
irb(main):007:1*   end
irb(main):008:0> end
=> :to_str
irb(main):009:0> f = IsNotString.new
irb(main):010:0> s = ""
irb(main):011:0> s << f
=> "bar"
```

But I would argue that `IsNotString` has a bug, because `to_s` does not return a string.

ext/liquid_c/vm.c
dylanahsmith: I moved the default case to the top of this switch statement, so it falls through to `case T_STRING` after coercing the object to a string.

macournoyer (approved):
LGTM

peterzhu2118 (approved):

ext/liquid_c/vm.c
dylanahsmith: Oh, I see.  Ruby does a `RB_TYPE_P(s, T_STRING)` check and instances of a string derived class also have the T_STRING object type, so it can be used for ruby functions like `rb_str_append` that expected a RString struct and won't be converted by `StringValue`.

Thanks for clarifying that for me. I've pushed a commit to get rid of the to_str conversion that shouldn't be necessary.

peterzhu2118 (approved):
Looks great!

Issue #80, opened 2020-10-09
Add profiling support


macournoyer:
Right now, only Ruby Liquid has profiling. Profiling Liquid templates in production triggers Ruby Liquid to render the template, not Liquid::C. This is leading us on the wrong paths RE what to optimize.

I need to look at how other VMs are doing profiling with low overhead.

dylanahsmith:
> Profiling Liquid templates in production

Is that something we are using?  Since it doesn't look like we are using the `profile: true` Liquid::Tempate.parse option.

> triggers Ruby Liquid to render the template, not Liquid::C. This is leading us on the wrong paths RE what to optimize.

Do you mean "us" as implementers of the liquid API?  Since I believe we use stackprof for profiling our code in production.  That will profile rendering with Liquid::C, although it won't show where the time is spent in the C code.  For that purpose, we will want a sampling C profiler, which works like stackprof, but gets the C stack trace periodically instead of the ruby stack trace.

For development of liquid-c on MacOS, I've added a c_profile rake task that works with Xcode's instruments command (https://github.com/Shopify/liquid-c/pull/62).

> I need to look at how other VMs are doing profiling with low overhead.

We should also allow template authors to profile their liquid code.

A sampling profile works better for a single tenant use case, where it can be used across renders and capture.  If we are just trying to profile a single render, then we may not collect enough samples.  If we increase the sampling frequency to collect more samples, then we add more significant overhead.  For local development, we render multiple times in a loop to collect enough samples with a sampling profiler, but naturally that redundant rendering is a significant overhead.

If we did want to implement a sampling profiler, then we could expose a stack of liquid source locations from the Liquid::Context that could be used by a sampler.

Another option would be to add instrumentation instructions to the VM code to collect timing information.  That is essentially how the ruby liquid profiler works, however, we could do that more efficiently from C code.  We would probably want to expose a different ruby API for it though.

macournoyer:
> Is that something we are using? Since it doesn't look like we are using the `profile: true` Liquid::Tempate.parse option.

[Yes](https://github.com/Shopify/storefront-renderer/blob/master/app/liquid/renderer.rb#L109), through the Theme Inspector that customers are using, and employees too.

[...]

PR #89, opened 2020-10-14
Add an `instrument` option to disable generation of native block bodies

macournoyer:
### Problem

As mentioned in https://github.com/Shopify/liquid-c/issues/80#issuecomment-708360946, currently all shops profiled through the Theme Inspector don't see the profiling of the section templates.

Screen Shot 2020-10-13 at 2 32 41 PM
(This should be 6 levels deep)

### Solution

We need a way to disable generation of native BlockBody::C when profiling will happen later, but don't want to profile yet.

The new option is to be used here: https://github.com/Shopify/storefront-renderer/blob/master/app/liquid/theme_render_context.rb#L265-L269

The name was to imply parse this for profiling (instrument it), but don't profile it. I imagined in the future, it will mean: generate byte-code that is instrumented. But opened to suggestions on the name.

### DIFF
--- a/lib/liquid/c.rb
+++ b/lib/liquid/c.rb
@@ -63,7 +63,10 @@ Liquid::ParseContext.class_eval do
   def disable_liquid_c_nodes
     # Liquid::Profiler exposes the internal parse tree that we don't want to build when
     # parsing with liquid-c, so consider liquid-c to be disabled when using it.
-    @disable_liquid_c_nodes ||= !Liquid::C.enabled || @template_options[:profile]
+    # Also, some templates are parsed before the profiler is running, on which case we
+    # provide the `instrument` option enabled the AST to be produced so the profiler can
+    # use it on future runs.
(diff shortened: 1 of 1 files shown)


macournoyer:
Setting `profile: true` for the section template's ParseContext cause a nil exception on line https://github.com/Shopify/liquid/blob/master/lib/liquid/profiler.rb#L90. I tried fixing that, but that didn't solve the issue of section profiling not showing up. So this seems like the simplest solution to buy some time before introducing profiling in the VM.

dylanahsmith:
This sounds like it might be an application specific issue, since it has to do with sections and not just the include/render tag.

It sounds like the intent of the `instrument: true` option is the same as the `profile: true` parse, which is to parse the template for profiling.

I think we could just name the option `:disable_liquid_c_nodes` which could be useful more generally, such as to use liquid parse tree.

dylanahsmith (approved):

PR #90, opened 2020-10-15
Call rb_gc_mark instead of rb_gc_mark_maybe

peterzhu2118:
`rb_gc_mark_locations` is usually used to mark the C stack, so it calls `rb_gc_mark_maybe` for each Ruby object in the memory region (see [here](https://github.com/ruby/ruby/blob/b68c22b3c6a80ebf96d446348bc8007b8a10f694/gc.c#L4957)). `rb_gc_mark_maybe` will only mark if the pointer is to a valid Ruby object. It checks that by performing binary search on the pages on the heap to ensure that the memory address is within a page. This is rather costly when we know that the VM stack only contains pointers to Ruby object so we could directly call `rb_gc_mark`.

### DIFF
--- a/ext/liquid_c/vm.c
+++ b/ext/liquid_c/vm.c
@@ -32,7 +32,11 @@ typedef struct vm {
 static void vm_mark(void *ptr)
 {
     vm_t *vm = ptr;
-    rb_gc_mark_locations((VALUE *)vm->stack.data, (VALUE *)vm->stack.data_end);
+
+    VALUE *stack_end = (VALUE *)vm->stack.data_end;
+    for (VALUE *obj_ptr = (VALUE *)vm->stack.data; obj_ptr < stack_end; obj_ptr++) {
+        rb_gc_mark(*obj_ptr);
+    }
     rb_gc_mark(vm->strainer);
     rb_gc_mark(vm->filter_methods);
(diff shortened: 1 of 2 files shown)


ext/liquid_c/vm.c
```diff
@@ -32,7 +32,11 @@ typedef struct vm {
 static void vm_mark(void *ptr)
 {
     vm_t *vm = ptr;
-    rb_gc_mark_locations((VALUE *)vm->stack.data, (VALUE *)vm->stack.data_end);
+
+    VALUE *stack_end = (VALUE *)vm->stack.data_end;
+    for (VALUE *obj_ptr = (VALUE *)vm->stack.data; obj_ptr < stack_end; obj_ptr++) {
+        rb_gc_mark(*obj_ptr);
+    }
```
dylanahsmith, line 39: We will probably want to extract this to a shared function to re-use in the `vm_assembler_gc_mark`.

dylanahsmith (approved):

ext/liquid_c/vm.c
peterzhu2118: I extracted it to `c_buffer.h`.

dylanahsmith (approved):

PR #92, opened 2020-10-15
Compile lax parsed variables and expressions from their ruby objects

dylanahsmith:
## Problem

Currently, we have a fallback to the lax parser that prevents compiling variables to liquid VM code.  Not only does this lose the render performance of the liquid VM, but it is another serialization concern for https://github.com/Shopify/liquid-c/issues/77.  We should get rid of the edge case of having to serialize or render a Liquid::Variable, Liquid::VariableLookup or Liquid::Expression.

## Solution

Start exposing VM assembler methods on the Liquid::C::BlockBody for compiling variables, variable lookups and expressions from ruby code.  Use these in the lax parse fallback for variables.

Note that the VM assembler methods exposed to ruby should be safe to use, such that we can't cause undefined behaviour from trying to render VM instructions added from ruby code.  As such, I have exposed separate vm_assembler_*_from_ruby functions that do additional checking for safety, such as `vm_assembler_require_stack_args` to make sure we will have the required number of values on the stack for the instruction being added.  In the future, we can extend this ruby VM compilation support to allow tags to be compiled from application code.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -68,7 +68,6 @@ static VALUE block_body_allocate(VALUE klass)
     vm_assembler_add_leave(&body->code);
     body->source = Qnil;
     body->render_score = 0;
-    body->parsing = false;
     body->blank = true;
     return obj;
 }
@@ -214,7 +213,7 @@ loop_break:
 
 static void ensure_not_parsing(block_body_t *body)
 {
-    if (body->parsing) {
+    if (body->code.parsing) {
         rb_raise(rb_eRuntimeError, "Liquid::C::BlockBody is in a incompletely parsed state");
     }
 }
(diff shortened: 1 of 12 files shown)


ext/liquid_c/vm_assembler.c
```diff
@@ -116,3 +139,103 @@ void vm_assembler_add_push_literal(vm_assembler_t *code, VALUE literal)
...
+    vm_assembler_require_stack_args(code, 2);
+    vm_assembler_add_new_int_range(code);
+}
+
+void vm_assembler_add_hash_new_from_ruby(vm_assembler_t *code, VALUE hash_size_obj)
+{
+    ensure_parsing(code);
+    unsigned int hash_size = NUM2USHORT(hash_size_obj);
+    if (hash_size > 255)
+        rb_enc_raise(utf8_encoding, cLiquidSyntaxError, "Hash literal has too many keys");
```
peterzhu2118, line 227: Can we move this check to `vm_assembler_add_hash_new` just like `vm_assembler_add_filter`?

ext/liquid_c/block.c
```diff
@@ -334,7 +332,7 @@ static VALUE block_body_nodelist(VALUE self)
                 break;
             }
 
-            case OP_POP_WRITE_VARIABLE:
+            case OP_RENDER_VARIABLE_RESCUE:
```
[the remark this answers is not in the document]
dylanahsmith: We just needed a marker for adding a variable placeholder to the nodelist.  I decided to make `OP_POP_WRITE` generic (rather than variable specific) before exposing it to ruby.

macournoyer (approved):
I really like the idea of compiling from Ruby, as it would make prototyping optimizations faster, and accessible to more people.

I know serialization needs to be solved first. Just a thought.

ext/liquid_c/vm_assembler.c
dylanahsmith: Yes, although, we still need the check in `keyword_arg_count > 255` in `try_variable_strict_parse` so that we can provide a better error message.

peterzhu2118 (approved):

PR #93, opened 2020-10-16
Fix error from calling FIX2INT on a fixnum that doesn't fit in an int

dylanahsmith:
## Problem

We were getting the following exception

```
bundler/gems/liquid-077bf2a409b9/lib/liquid/block.rb:76:in `parse': integer 3155695200 too big to convert to `int' (RangeError)
```

which seems to be from trying to convert a fixnum into an int in `vm_assembler_add_push_fixnum`.

## Solution

The fix was to simply convert a fixnum into a long instead of an integer.

The expression tests were testing constant fixnums, which don't actually get compiled to optimize for those simple expressions and avoid a possible incompatibility problem.  So I added tests for actually compiling those constants, including a regression test in `test_push_large_fixnum`.

### DIFF
--- a/ext/liquid_c/vm_assembler.c
+++ b/ext/liquid_c/vm_assembler.c
@@ -85,7 +85,7 @@ void vm_assembler_add_write_node(vm_assembler_t *code, VALUE node)
 
 void vm_assembler_add_push_fixnum(vm_assembler_t *code, VALUE num)
 {
-    int x = FIX2INT(num);
+    long x = FIX2LONG(num);
     if (x >= INT8_MIN && x <= INT8_MAX) {
         vm_assembler_add_push_int8(code, x);
     } else if (x >= INT16_MIN && x <= INT16_MAX) {
--- a/test/unit/expression_test.rb
+++ b/test/unit/expression_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 
 class ExpressionTest < MiniTest::Test
-  def test_literals
+  def test_constant_literals
     assert_equal true, Liquid::C::Expression.strict_parse('true')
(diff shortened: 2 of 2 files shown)


peterzhu2118 (approved):

ext/liquid_c/vm_assembler.c
```diff
@@ -85,7 +85,7 @@ void vm_assembler_add_write_node(vm_assembler_t *code, VALUE node)
 
 void vm_assembler_add_push_fixnum(vm_assembler_t *code, VALUE num)
 {
-    int x = FIX2INT(num);
+    long x = FIX2LONG(num);
```
pushrax, line 88:
```c
static inline int
RB_FIX2INT(VALUE x)
{
    long ret;

    if /* constexpr */ (sizeof(int) < sizeof(long)) {
        ret = rb_fix2int(x);
    }
    else {
        ret = RB_FIX2LONG(x);
    }

    return RBIMPL_CAST((int)ret);
}

```

I guess `sizeof(int) < sizeof(long)`?

dylanahsmith: `sizeof(int) < sizeof(long)` is true for our purposes on a 64-bit system, although they are normally both 4 on a 32-bit system.

pushrax: Ah right, it's not that int is small, it's that long is actually long long. So I guess the errors is from

```c
#if SIZEOF_INT < SIZEOF_LONG
static void
check_int(long num)
{
    if ((long)(int)num != num) {
	rb_out_of_int(num);
    }
}
```

dylanahsmith: exactly

PR #94, opened 2020-10-16
Fix stack size calculations for key lookup instructions

dylanahsmith:
## Problem

I tried adding an assertion in `internal_expression_parse` to make sure the expression evaluates to exactly one value and got an assertion error when running the tests.  It looks like the lookup key instructions were overestimated the stack size, causing more space to be reserved on the stack than needed.

## Solution

Add the assertion along with fixes for the stack size adjustments in the add lookup key instruction functions.

### DIFF
--- a/ext/liquid_c/vm_assembler.h
+++ b/ext/liquid_c/vm_assembler.h
@@ -65,6 +65,12 @@ static inline void vm_assembler_increment_stack_size(vm_assembler_t *code, size_
         code->max_stack_size = code->stack_size;
 }
 
+static inline void vm_assembler_reserve_stack_size(vm_assembler_t *code, size_t amount)
+{
+    vm_assembler_increment_stack_size(code, amount);
+    code->stack_size -= amount;
+}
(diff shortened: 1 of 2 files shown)


ext/liquid_c/vm_assembler.h
```diff
@@ -158,20 +164,20 @@ static inline void vm_assembler_add_find_variable(vm_assembler_t *code)
 
 static inline void vm_assembler_add_lookup_const_key(vm_assembler_t *code, VALUE key)
 {
-    vm_assembler_increment_stack_size(code, 1);
+    vm_assembler_reserve_stack_size(code, 1); // push 1, pop 2, push 1
```
dylanahsmith, line 167: This looks a bit weird here and in `vm_assembler_add_lookup_command`, because we push the value on the stack to fall-through to the case for `OP_LOOKUP_KEY` in the VM switch statement.

peterzhu2118 (approved):

PR #95, opened 2020-10-16
Store some Liquid::Context instance variables in the VM

peterzhu2118:
Store the `@environments`, `@static_environments`, `@scopes` variables of `Liquid::Context` in the VM (inside the `context_t` struct). This means we don't need to fetch these instance variables every time we call `Liquid::Context#find_variable`.

I didn't store the `@strict_variables` variable because it's a writable attribute (so it could change from initialization).

### DIFF
--- a/ext/liquid_c/vm.c
+++ b/ext/liquid_c/vm.c
@@ -22,11 +22,6 @@ static void vm_mark(void *ptr)
     vm_t *vm = ptr;
 
     c_buffer_rb_gc_mark(&vm->stack);
-    rb_gc_mark(vm->strainer);
-    rb_gc_mark(vm->filter_methods);
-    rb_gc_mark(vm->interrupts);
-    rb_gc_mark(vm->resource_limits_obj);
-    rb_gc_mark(vm->global_filter);
     context_mark(&vm->context);
 }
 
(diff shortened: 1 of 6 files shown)


ext/liquid_c/vm.c
```diff
@@ -67,25 +50,29 @@ static VALUE vm_internal_new(VALUE context)
...
-    vm->resource_limits_obj = rb_ivar_get(context, id_ivar_resource_limits);;
-    ResourceLimits_Get_Struct(vm->resource_limits_obj, vm->resource_limits);
+    vm->context.resource_limits_obj = rb_ivar_get(context, id_ivar_resource_limits);;
+    ResourceLimits_Get_Struct(vm->context.resource_limits_obj, vm->context.resource_limits);
 
-    vm->strict_filters = RTEST(rb_funcall(context, id_strict_filters, 0));
-    vm->global_filter = rb_funcall(context, id_global_filter, 0);
+    vm->context.strict_variables = false;
+    vm->context.strict_filters = RTEST(rb_funcall(context, id_strict_filters, 0));
+    vm->context.global_filter = rb_funcall(context, id_global_filter, 0);
```
dylanahsmith, line 67: This initialization could be moved into context_internal_init to keep the code coupled to Liquid::Context together.

dylanahsmith (approved):

PR #97, opened 2020-10-21
Compile using -O0 in debugging

peterzhu2118:
Compile using `-O0` in debugging to prevent optimizations from erasing functions and variables that can no longer be accessed in a debugger.

`inline` functions are only possibly inlined at `-O2` and higher. Without the `static` keyword, these inline functions are supposed to exist externally. Which means that they fail to run when they are not inlined (since it can't find these functions that don't exist externally). The solution is to add the `static` keyword to the `inline` functions. Many thanks to @dylanahsmith for helping me debug this!

See this SO answer for more info: https://stackoverflow.com/a/41218504/3878329

### DIFF
--- a/ext/liquid_c/c_buffer.h
+++ b/ext/liquid_c/c_buffer.h
@@ -9,28 +9,28 @@ typedef struct c_buffer {
     uint8_t *capacity_end;
 } c_buffer_t;
 
-inline c_buffer_t c_buffer_init()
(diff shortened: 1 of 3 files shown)


dylanahsmith:
Looks like it will be more important to use `DEBUG=false bundle exec rake clean compile` before profiling.

dylanahsmith (approved):

PR #99, opened 2020-10-23
Fix ambiguous first argument warnings when running the unit tests

dylanahsmith:
## Problem

I'm seeing the following warnings when running the unit tests

```
/Users/dylants/src/liquid-c/test/unit/expression_test.rb:27: warning: ambiguous first argument; put parentheses or a space even after `-' operator
/Users/dylants/src/liquid-c/test/unit/expression_test.rb:32: warning: ambiguous first argument; put parentheses or a space even after `-' operator
/Users/dylants/src/liquid-c/test/unit/expression_test.rb:34: warning: ambiguous first argument; put parentheses or a space even after `-' operator
/Users/dylants/src/liquid-c/test/unit/expression_test.rb:39: warning: ambiguous first argument; put parentheses or a space even after `-' operator
/Users/dylants/src/liquid-c/test/unit/expression_test.rb:51: warning: ambiguous first argument; put parentheses or a space even after `-' operator
```

which are coming from lines like `assert_equal -128, compile_and_eval('-128')`

## Solution

Use parenthesis for the `assert_equal` calls in that file to avoid these warnings and to make it harder to introduce these by convention.

### DIFF
--- a/test/unit/expression_test.rb
+++ b/test/unit/expression_test.rb
@@ -2,58 +2,58 @@ require 'test_helper'
 
 class ExpressionTest < MiniTest::Test
   def test_constant_literals
-    assert_equal true, Liquid::C::Expression.strict_parse('true')
-    assert_equal false, Liquid::C::Expression.strict_parse('false')
-    assert_nil Liquid::C::Expression.strict_parse('nil')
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):
We should consider setting up rubocop to catch issues like these

PR #100, opened 2020-10-23
Set @line_number directly on the Liquid::ParseContext

dylanahsmith:
`Liquid::ParseContext` uses an `attr_accessor` for the line number.  Right now liquid-c is calling `Liquid::ParseContext#line_number=` to set that attribute, but it would be faster to just call `rb_ivar_set`.  Also, this will let me re-use the same ID for re-using with `rb_attr_get`, which I will need for Liquid::Assign.compile in https://github.com/Shopify/liquid-c/pull/96

### DIFF
--- a/ext/liquid_c/liquid.c
+++ b/ext/liquid_c/liquid.c
@@ -16,6 +16,7 @@ ID id_to_liquid;
 ID id_to_s;
 ID id_call;
 ID id_compile_evaluate;
+ID id_ivar_line_number;
 
 VALUE mLiquid, mLiquidC, cLiquidVariable, cLiquidTemplate, cLiquidBlockBody;
 VALUE cLiquidVariableLookup, cLiquidRangeLookup;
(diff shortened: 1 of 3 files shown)


ext/liquid_c/liquid.c
```diff
@@ -36,6 +37,7 @@ void Init_liquid_c(void)
     id_to_s = rb_intern("to_s");
     id_call = rb_intern("call");
     id_compile_evaluate = rb_intern("compile_evaluate");
+    id_ivar_line_number = rb_intern("@line_number");
```
peterzhu2118, line 40: Is there a reason `id_ivar_line_number` is declared in `liquid.c` rather than `block.c`? Is it/will it be used anywhere else?

dylanahsmith: > this will let me re-use the same ID for re-using with `rb_attr_get`, which I will need for Liquid::Assign.compile in #96

peterzhu2118 (approved):

PR #101, opened 2020-10-23
Add rubocop checks

peterzhu2118:
Adds rubocop checks. Uses the defaults in [Shopify/ruby-style-guide](https://github.com/Shopify/ruby-style-guide). Most of the changes were autocorrected by rubocop.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,14 @@
+# frozen_string_literal: true
 source 'https://rubygems.org'
 
 gemspec
 
 gem 'liquid', github: 'Shopify/liquid', ref: 'master'
 
-
 group :test do
+  gem 'rubocop', '~> 0.93.1', require: false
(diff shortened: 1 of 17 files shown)


Gemfile
```diff
@@ -1,11 +1,14 @@
...
 
 gemspec
 
 gem 'liquid', github: 'Shopify/liquid', ref: 'master'
 
-
 group :test do
+  gem 'rubocop', require: false
+  gem 'rubocop-performance', require: false
+  gem 'rubocop-shopify', require: false
```
dylanahsmith, line 11: Shouldn't we specify a version?  Otherwise, I would expect that updates will cause CI to start failing for this library.

peterzhu2118: Updated. I also noticed that Shopify/ruby-style-guide does not yet support Rubocop 1.0.

dylanahsmith (approved):

Issue #78, opened 2020-10-08
Compile nested block bodies into their parent block body object


dylanahsmith:
Currently, the library writes liquid VM code into separate growable buffers (in vm_assembler_t) for each Liquid::C::BlockBody object.  However, for block body serialization and deserialization (https://github.com/Shopify/liquid-c/issues/77) we will want to share an immutable memory allocation across block bodies, so block bodies will need offsets into that shared memory for execution.

Since block bodies will be connected in this way for execution, we may also want to make parsing more efficient by re-using a `vm_assembler_t` for a block body once that block body has finished being compiled (e.g. when its parent block tag is finished compiling), by not embedding it in the `block_body_t` struct.  That way we could save on the number of memory allocations for these vm_assembler_t buffers and will have less `vm_assembler_t` buffers to combine at the end of parsing / compilation.  For instance, we could have one `vm_assembler_t` per-level of nesting in the template.

dylanahsmith:
I think this was fixed in https://github.com/Shopify/liquid-c/pull/102

PR #102, opened 2020-10-26
Write block body to a shared buffer after compilation

peterzhu2118:
Based on #86.

Write `BlockBody` to a buffer stored in the `ParseContext` (as an instance variable called `@document_body`) after parsing has completed. We know parsing has completed when `BlockBody#freeze` is called ([introduced in (branch `pz-freeze-block` of Shopify/liquid](https://github.com/Shopify/liquid/tree/pz-freeze-block)). Freezing the `BlockBody` , it writes the body to the buffer and prevents further changes to body (i.e. not allowed to add more nodes).

I've introduced a new struct called `internal_block_body_t` that holds the intermediate state (before `BlockBody#freeze` is called) and the state after compilation (pointers to the buffer and offset in the buffer). As the name suggests, this struct is used to keep track of the internal state of `BlockBody`. `block_body_t` is the struct used to store the actual data structure stored in the buffer and available publicly.

Currently, the `vm_assembler_t` are not shared, but I will implement that in a later commit.

See #78.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -7,6 +7,7 @@
 #include "variable.h"
 #include "context.h"
 #include "document_body.h"
+#include "parse_context.h"
 #include "vm_assembler.h"
 #include <stdio.h>
 
@@ -33,7 +34,7 @@ typedef struct parse_context {
     VALUE ruby_obj;
 } parse_context_t;
 
-typedef struct internal_block_body {
+typedef struct block_body {
     bool compiled;
     VALUE obj;
 
@@ -42,43 +43,54 @@ typedef struct internal_block_body {
             VALUE document_body_obj;
             c_buffer_t *buffer;
             size_t offset;
+            VALUE constants;
+            VALUE nodelist;
         } compiled;
         struct {
-            VALUE source;
+            VALUE parse_context;
             bool blank;
             bool root;
             int render_score;
             vm_assembler_t code;
         } intermediate;
     } as;
-} internal_block_body_t;
+} block_body_t;
 
-static block_body_t *block_body_ptr(const internal_block_body_t *body)
+static void ensure_body_compiled(const block_body_t *body)
 {
-    return ((block_body_t *)(body->as.compiled.buffer->data + body->as.compiled.offset));
+    if (!body->compiled) {
+        rb_raise(rb_eRuntimeError, "Liquid::C::BlockBody has not been compiled");
+    }
+}
+
+static block_body_header_t *block_body_header_ptr(const block_body_t *body)
+{
+    ensure_body_compiled(body);
+    return ((block_body_header_t *)(body->as.compiled.buffer->data + body->as.compiled.offset));
+}
+
+static const VALUE *block_body_constants_ptr(const block_body_t *body)
+{
+    block_body_header_t *header = block_body_header_ptr(body);
+    return RARRAY_PTR(body->as.compiled.constants) + header->constants_offset;
 }
 
 static void block_body_mark(void *ptr)
 {
-    internal_block_body_t *body = ptr;
+    block_body_t *body = ptr;
     if (body->compiled) {
         rb_gc_mark(body->as.compiled.document_body_obj);
-        block_body_t *block_body = block_body_ptr(body);
-        VALUE *start = block_body_constants_ptr(block_body);
-        assert(block_body->constants_bytes % sizeof(VALUE *) == 0);
-        size_t len = block_body->constants_bytes / sizeof(VALUE *);
-        for (size_t i = 0; i < len; i++) {
-            rb_gc_mark(start[i]);
-        }
+        rb_gc_mark(body->as.compiled.constants);
+        rb_gc_mark(body->as.compiled.nodelist);
     } else {
-        rb_gc_mark(body->as.intermediate.source);
+        rb_gc_mark(body->as.intermediate.parse_context);
         vm_assembler_gc_mark(&body->as.intermediate.code);
     }
 }
 
 static void block_body_free(void *ptr)
 {
-    internal_block_body_t *body = ptr;
+    block_body_t *body = ptr;
     if (!body->compiled) {
         vm_assembler_free(&body->as.intermediate.code);
     }
@@ -87,13 +99,12 @@ static void block_body_free(void *ptr)
 
 static size_t block_body_memsize(const void *ptr)
 {
-    const internal_block_body_t *body = ptr;
+    const block_body_t *body = ptr;
     if (!ptr) return 0;
     if (body->compiled) {
-        block_body_t *block_body = block_body_ptr(body);
-        return sizeof(internal_block_body_t) + block_body->instructions_bytes + block_body->constants_bytes;
+        return sizeof(block_body_t);
     } else {
-        return sizeof(internal_block_body_t) + vm_assembler_alloc_memsize(&body->as.intermediate.code);
+        return sizeof(block_body_t) + vm_assembler_alloc_memsize(&body->as.intermediate.code);
     }
 }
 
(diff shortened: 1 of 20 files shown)


ext/liquid_c/block.c
```diff
@@ -30,25 +33,68 @@ typedef struct parse_context {
...
 
 static size_t block_body_memsize(const void *ptr)
 {
-    const block_body_t *body = ptr;
+    const internal_block_body_t *body = ptr;
     if (!ptr) return 0;
-    return sizeof(block_body_t) + vm_assembler_alloc_memsize(&body->code);
+    if (body->compiled) {
+        block_body_t *block_body = block_body_ptr(body);
+        return sizeof(internal_block_body_t) + block_body->instructions_bytes + block_body->constants_bytes;
```
dylanahsmith, line 94: Isn't the instruction and constant bytes a part of the document body at this point?  So I think you should count that memory in the document body object.

ext/liquid_c/context.c
```diff
@@ -131,6 +132,27 @@ VALUE context_filtering_p(VALUE self)
     return liquid_vm_filtering(self) ? Qtrue : Qfalse;
 }
 
+bool context_init_document_body(VALUE self)
```
dylanahsmith, line 135: This makes it seem like it is referring to the `Liquid::Context` but it is actually referring to the `Liquid::ParseContext`.  To avoid confusion, we should use `parse_context` instead of shortening it to `context` in names referring to the `ParseContext` and we should also use a separate C file if we want a place for C code related to the parse context.

ext/liquid_c/context.c
```diff
@@ -142,6 +164,7 @@ void init_liquid_context()
     id_ivar_environments = rb_intern("@environments");
     id_ivar_static_environments = rb_intern("@static_environments");
     id_ivar_strict_variables = rb_intern("@strict_variables");
+    id_ivar_document_body = rb_intern("@document_body");
```
dylanahsmith, line 167: Are we OK with ruby code being able to set or remove this instance variable?  If not, then we could keep it internal to C code by removing the `@` prefix to the instance variable name.

ext/liquid_c/block.h
```diff
@@ -1,17 +1,26 @@
...
 #define LIQUID_BLOCK_H
 
-#include "vm_assembler.h"
-
 typedef struct block_body {
-    VALUE obj;
-    vm_assembler_t code;
-    VALUE source; // hold a reference to the ruby object that OP_WRITE_RAW points to
+    size_t instructions_offset;
+    size_t instructions_bytes;
```
dylanahsmith, line 6: It seems like this is the serialized form for the block body, in which case we want to avoid using `size_t` values, since that seems like unnecessary coupling to the CPU architecture that it is compiled on.  If we only support a 24-bit length source string, then we can safely use `uint32_t` for offsets and lengths into the serialized code, assuming a byte of source can't compile into more than 256 bytes of compiled code.

ext/liquid_c/vm.c
```diff
@@ -170,7 +171,7 @@ static void write_obj(VALUE output, VALUE obj)
 static inline void vm_stack_push(vm_t *vm, VALUE value)
 {
     VALUE *stack_ptr = (VALUE *)vm->stack.data_end;
-    assert(stack_ptr < (VALUE *)vm->stack.capacity_end);
+    assert(stack_ptr <= (VALUE *)vm->stack.capacity_end);
```
dylanahsmith, line 174: We shouldn't be writing to `vm->stack.capacity_end`, since that would be past the end of the allocation.

dylanahsmith (changes requested):

ext/liquid_c/document_body.c
```diff
@@ -0,0 +1,80 @@
...
+        .constants_offset = sizeof(block_body_t) + instructions_bytes,
+        .constants_bytes = constants_bytes,
+        .blank = blank,
+        .render_score = render_score,
+        .max_stack_size = code->max_stack_size
+    };
+
+    c_buffer_write(&body->buffer, &buf_block_body, sizeof(block_body_t));
+    c_buffer_concat(&body->buffer, &code->instructions);
+    c_buffer_concat(&body->buffer, &code->constants);
```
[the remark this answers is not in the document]
peterzhu2118: Do we want to replace the `constants` buffer in `vm_assembler_t` with a Ruby array or should we create a Ruby array when writing it to the document body?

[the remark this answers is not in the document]
dylanahsmith: We don't need to switch it in `vm_assembler_t` for now if we aren't using that for execution, since we will need to copy it into the ruby array either way.  However, feel free to change it in the `vm_assembler_t` if that makes things easier.

ext/liquid_c/block.h
peterzhu2118: I think we should use `uint64_t` for the offset because since we are writing every block body to this shared buffer, the buffer size could overflow a `uint32_t` (e.g. if each file compiled `2^24` bytes of instructions and we have `2^8=256` files).

dylanahsmith: A document body would map to only a single file.  Each partial would have its own document body.  However, if we did decide to combine them in the future, then we would have to worry more about overflows.

ext/liquid_c/document_body.c
```diff
@@ -0,0 +1,86 @@
...
+        .instructions_offset = (uint32_t)sizeof(block_body_header_t),
+        .instructions_bytes = (uint32_t)c_buffer_size(&code->instructions),
+        .constants_offset = (uint32_t)RARRAY_LEN(body->constants),
+        .constants_len = (uint32_t)c_buffer_size(&code->constants) / sizeof(VALUE *),
+        .blank = blank,
+        .render_score = render_score,
+        .max_stack_size = code->max_stack_size
+    };
+
+    c_buffer_write(&body->buffer, &buf_block_body, sizeof(block_body_header_t));
```
dylanahsmith, line 72: I think we could use the `c_buffer_extend_for_write` function from https://github.com/Shopify/liquid-c/pull/112 to write these values directly to the buffer, in order to avoid having to copy them.

```suggestion
    block_body_header_t *buf_block_body = c_buffer_extend_for_write(&body->buffer, sizeof(block_body_header_t));
    *buf_block_body = {
        .instructions_offset = (uint32_t)sizeof(block_body_header_t),
        .instructions_bytes = (uint32_t)c_buffer_size(&code->instructions),
        .constants_offset = (uint32_t)RARRAY_LEN(body->constants),
        .constants_len = (uint32_t)c_buffer_size(&code->constants) / sizeof(VALUE *),
        .blank = blank,
        .render_score = render_score,
        .max_stack_size = code->max_stack_size
    };
```

dylanahsmith, line 72: I think we should also align the struct to avoid problems from unaligned memory access.  We could add the following function to c_buffer.c to help with that

```c
void c_buffer_zero_pad_for_alignment(c_buffer_t *buffer, size_t alignment)
{
    size_t unaligned_bytes = c_buffer_size(buffer) % alignment;
    if (unaligned_bytes) {
        size_t pad_size = alignment - unaligned_bytes;
        uint8_t *padding = c_buffer_extend_for_write(buffer, pad_size);
        memset(padding, 0, pad_size);
    }
}
```

then we can use the C11 `alignof` macro (from `stdalign.h`) to pass the alignment value

```c
    c_buffer_zero_pad_for_alignment(&body->buffer, alignof(block_body_header_t));
```

ext/liquid_c/block.h
```diff
@@ -1,17 +1,46 @@
...
-    vm_assembler_t code;
+typedef struct block_body_header {
+    uint32_t instructions_offset;
+    uint32_t instructions_bytes;
+    uint32_t constants_offset;
+    uint32_t constants_len;
     bool blank;
     int render_score;
-    VALUE nodelist;
+    size_t max_stack_size;
```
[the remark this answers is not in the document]
peterzhu2118: To do this, I think the `stack_size` variables in `vm_assembler_t` should also change to `uint32_t`. I think this should be fine. WDYT?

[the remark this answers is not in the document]
dylanahsmith: Good point.  I think that should be fine for now, although we might need to be more careful in the future with allowing ruby code to assemble VM code, so we could use uint64_t if we want to be more conservative about what we allow.  However, I would like to be more specific about the serialization format than using `size_t`.

dylanahsmith (approved):

PR #103, opened 2020-10-26
Allow up to 32-bits for source file size

peterzhu2118:
#81 introduced a restriction that enforces source files to be 24-bits in length (16MiB). This doesn't seem to be enough sometimes. This PR changes the restrictions to be 32-bits in length (4GiB) which should be enough.


peterzhu2118:
Closing because #104 is merged.

PR #104, opened 2020-10-27
Fallback to not using the liquid VM if the source is too large

dylanahsmith:
## Problem

It turns out that we do have usage of liquid with a larger than 24-bit length source strings.

We could workaround this in the application by using the `disable_liquid_c_nodes` parse option.  However, it doesn't seem like the application should have to be aware of this limit.

## Solution

Add a fallback in the ruby patch for Liquid::Tokenizer.new to the ruby tokenizer if the source is too large to use the liquid-c tokenizer.  We want to propagate this to the Liquid::ParseContext, so I did this in a Liquid::Document.parse method wrapper.

### DIFF
--- a/test/unit/tokenizer_test.rb
+++ b/test/unit/tokenizer_test.rb
@@ -64,11 +64,28 @@ class TokenizerTest < Minitest::Test
   end
 
   def test_source_too_large
-    err = assert_raises(ArgumentError) do
-      tokenize("a" * 2**24)
-    end
+    too_large_source = "a" * 2**24
+    max_length_source = too_large_source.chop
 
+    # C safety check
+    err = assert_raises(ArgumentError) do
+      Liquid::C::Tokenizer.new(too_large_source, 1, false)
+    end
     assert_match(/Source too large, max \d+ bytes/, err.message)
+
+    # ruby patch fallback
+    liquid_c_tokenizer = Liquid::Tokenizer.new(max_length_source)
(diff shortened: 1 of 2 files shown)


test/unit/tokenizer_test.rb
```diff
@@ -64,11 +64,28 @@ def test_non_utf8_compatible_source
...
+    fallback_tokenizer = Liquid::Tokenizer.new(too_large_source)
+    assert_instance_of(Liquid::Tokenizer, fallback_tokenizer)
+
+    # Document.parse patch parse context update
+    parse_context = Liquid::ParseContext.new
+    refute(parse_context.send(:disable_liquid_c_nodes))
+    Liquid::Document.parse(liquid_c_tokenizer, parse_context)
+    refute(parse_context.send(:disable_liquid_c_nodes))
+    Liquid::Document.parse(fallback_tokenizer, parse_context)
+    assert(true, parse_context.send(:disable_liquid_c_nodes))
```
peterzhu2118, line 88: I think you meant:
```suggestion
    assert(parse_context.send(:disable_liquid_c_nodes))
```
Otherwise I think it's a trivial assertion.

dylanahsmith: Oops.  This was meant to be `assert_equal`.  It is just the falsey case where I didn't want to bother with the difference between `nil` and `false`.  Pushed a fix

samdoiron (approved):

peterzhu2118 (approved):

PR #110, opened 2020-10-29
Add disassemble method to Liquid::C::BlockBody and Liquid::C::Expression

dylanahsmith:
## Problem

After compilation, Liquid::C::BlockBody and Liquid::C::Expression objects are completely opaque.  This makes it hard to inspect the state of these objects for debugging or to test compilation to specialized instructions when a more general one would have the same semantics.

## Solution

Add disassemble methods that return a string that can be printed with `puts` for debugging (e.g. `puts template.root.body.disassemble`) or used as a test assertion as the included unit tests demonstrate.

### DIFF
--- a/ext/liquid_c/vm_assembler.c
+++ b/ext/liquid_c/vm_assembler.c
@@ -126,8 +126,7 @@ VALUE vm_assembler_disassemble(vm_assembler_t *code)
 
             case OP_PUSH_INT16:
             {
-                int num = *(int8_t *)ip++;
-                num = (num << 8) | ip[1];
+                int num = (ip[1] << 8) | ip[2];
                 rb_str_catf(output, "push_int16(%u)\n", num);
                 break;
             }
--- a/ext/liquid_c/vm_assembler.c
+++ b/ext/liquid_c/vm_assembler.c
@@ -172,7 +172,7 @@ VALUE vm_assembler_disassemble(vm_assembler_t *code)
                 break;
 
             default:
-                rb_str_catf(output, "<%d>\n", ip[0]);
+                rb_str_catf(output, "<opcode number %d disassembly not implemented>\n", ip[0]);
                 break;
         }
         liquid_vm_next_instruction(&ip, &const_ptr);
--- a/ext/liquid_c/vm_assembler.c
+++ b/ext/liquid_c/vm_assembler.c
@@ -1,6 +1,7 @@
 #include "liquid.h"
 #include "vm_assembler.h"
 #include "expression.h"
+#include "vm.h"
 
 void vm_assembler_init(vm_assembler_t *code)
 {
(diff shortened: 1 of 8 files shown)


ext/liquid_c/vm_assembler.c
```diff
@@ -73,6 +74,113 @@ void vm_assembler_gc_mark(vm_assembler_t *code)
...
+                rb_str_catf(output, "render_variable_rescue(line_number: %u)\n", line_number);
+                break;
+            }
+
+            case OP_WRITE_RAW:
+            {
+                const char *text = (const char *)const_ptr[0];
+                size_t size = const_ptr[1];
+                VALUE string = rb_enc_str_new(text, size, utf8_encoding);
+                rb_str_catf(output, "write_raw(%+"PRIsVALUE")\n", string);
```
peterzhu2118, line 147: Can we do this?
```suggestion
                rb_str_catf(output, "write_raw(\"%.*s\")\n", (int)size, text);
```
It would avoid allocating a string.

ext/liquid_c/vm_assembler.c
```diff
@@ -73,6 +74,113 @@ void vm_assembler_gc_mark(vm_assembler_t *code)
...
+                break;
+
+            case OP_PUSH_INT8:
+                rb_str_catf(output, "push_int8(%u)\n", ip[1]);
+                break;
+
+            case OP_PUSH_INT16:
+            {
+                int num = *(int8_t *)ip++;
+                num = (num << 8) | ip[1];
```
peterzhu2118, line 130: `OP_PUSH_INT16` is broken.
```suggestion
                int num = (ip[1] << 8) | ip[2];
```

ext/liquid_c/vm_assembler.c
```diff
@@ -73,6 +74,113 @@ void vm_assembler_gc_mark(vm_assembler_t *code)
...
+            case OP_LOOKUP_COMMAND:
+                rb_str_catf(output, "lookup_command(%+"PRIsVALUE")\n", const_ptr[0]);
+                break;
+
+            case OP_FILTER:
+                rb_str_catf(output, "filter(name: %+"PRIsVALUE", num_args: %u)\n", const_ptr[0], ip[1]);
+                break;
+
+            default:
+                rb_str_catf(output, "<%d>\n", ip[0]);
```
peterzhu2118, line 176: Should we output a more obvious error message here?

dylanahsmith: It isn't exactly an error, since it is safe to show the opcode number if the opcode is handled in `liquid_vm_next_instruction`.  However, you're right that it wasn't clear why just an number is shown like that.  I pushed a commit to change it to `rb_str_catf(output, "<opcode number %d disassembly not implemented>\n", ip[0]);`

ext/liquid_c/vm_assembler.c
dylanahsmith: You're right.  I pushed a commit to fix it along with a test for it.

ext/liquid_c/vm_assembler.c
dylanahsmith: I'm not really concerned about the performance of disassemble, since it is provided for testing and debugging.  Your suggested change wouldn't escape special characters, such as newlines or quote characters.

peterzhu2118: Ah yeah I didn't think about newlines/Special characters. Makes sense, thanks for clarifying!

peterzhu2118 (approved):

PR #111, opened 2020-10-29
Strict parse Liquid::Variable.new objects to a Liquid::C::Expression

dylanahsmith:
This solves the problem that https://github.com/Shopify/liquid-c/pull/96 intended to solve in a simpler way, so that PR can be focused on introducing support for compiling tags to VM code.  Instead, this PR compiles the variable expression and filtering into a `Liquid::C::Expression` object that is stored in the `Liquid::Variable` object's `@name` instance variable, which will be evaluated on render.

## Benchmark

The liquid benchmark only uses the assign tag `{% assign article = pages.frontpage %}` and doesn't use the echo tag, so the benchmark is mostly for unaffected code.  However, I wrote a micro-benchmark that shows the performance improvement

```ruby
# frozen_string_literal: true

require 'bundler/setup'
require 'benchmark/ips'
require 'liquid/c'

PARSE_CONTEXT = Liquid::ParseContext.new
CONTEXT = Liquid::Context.new('x' => 1, 'y' => 2)

def parse
  Liquid::Variable.new("x | plus: y", PARSE_CONTEXT)
end

Benchmark.ips do |x|
  x.report("parse") { parse }
  variable = parse
  x.report("render") { variable.render(CONTEXT) }
  x.report("parse & render") { parse.render(CONTEXT) }
end
```

result on master

```
               parse     88.381k (± 3.2%) i/s -    445.068k in   5.041305s
              render    510.144k (± 2.3%) i/s -      2.577M in   5.053979s
      parse & render     67.412k (± 4.0%) i/s -    342.414k in   5.088448s
```

result on this branch

```
               parse    631.565k (± 3.0%) i/s -      3.160M in   5.009016s
              render    961.840k (± 4.0%) i/s -      4.858M in   5.060674s
      parse & render    355.364k (± 3.3%) i/s -      1.790M in   5.041629s
```

### DIFF
--- a/ext/liquid_c/variable.c
+++ b/ext/liquid_c/variable.c
@@ -13,10 +13,10 @@ static VALUE try_variable_strict_parse(VALUE uncast_args)
     init_parser(&p, parse_args->markup, parse_args->markup_end);
     vm_assembler_t *code = &parse_args->body->code;
 
-    if (p.cur.type == TOKEN_EOS)
+    if (p.cur.type == TOKEN_EOS) {
+        vm_assembler_add_push_nil(code);
         return Qnil;
-
-    vm_assembler_add_render_variable_rescue(code, parse_args->line_number);
+    }
 
     parse_and_compile_expression(&p, code);
 
@@ -70,8 +70,6 @@ static VALUE try_variable_strict_parse(VALUE uncast_args)
         vm_assembler_add_filter(code, filter_name, arg_count);
     }
 
-    vm_assembler_add_pop_write(code);
-
     parser_must_consume(&p, TOKEN_EOS);
 
     return Qnil;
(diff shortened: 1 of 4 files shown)


ext/liquid_c/variable.c
```diff
@@ -128,8 +130,54 @@ void internal_variable_parse(variable_parse_args_t *parse_args)
...
+
+    variable_parse_args_t parse_args = {
+        .markup = RSTRING_PTR(markup),
+        .markup_end = RSTRING_END(markup),
+        .code = &expression->code,
+        .code_obj = expression_obj,
+        .parse_context = parse_context,
+    };
+    try_variable_strict_parse((VALUE)&parse_args);
+    RB_GC_GUARD(markup);
```
peterzhu2118, line 159: I'm not sure why this is needed. How is `markup` at risk of being GC'd?

dylanahsmith: isn't needed before `try_variable_strict_parse` is called, so the `markup` local variable could get re-used by the compiler and remove the reference to the object from the C stack.  If the only reference to the `markup` object were passed into this method, then in theory, it seems like the GC could end up cleanup the string before we are done using it.

It probably isn't needed in practice, but it makes the code more obviously correct.

peterzhu2118: Thanks for clarifying!

peterzhu2118 (approved):

macournoyer (approved):
🤯 

Just unsure about that push_nil at the end.

ext/liquid_c/variable.c
```diff
@@ -6,17 +6,23 @@
...
 {
     variable_parse_args_t *parse_args = (void *)uncast_args;
     parser_t p;
     init_parser(&p, parse_args->markup, parse_args->markup_end);
-    vm_assembler_t *code = &parse_args->body->code;
+    vm_assembler_t *code = parse_args->code;
 
-    if (p.cur.type == TOKEN_EOS)
+    if (p.cur.type == TOKEN_EOS) {
+        vm_assembler_add_push_nil(code);
```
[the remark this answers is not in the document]
dylanahsmith: Yeah, we need something left on the stack since it will be followed by a pop to get the return value

PR #117, opened 2020-11-03
Fix Liquid::C::Tokenizer compatibility for liquid tag

dylanahsmith:
cc @wizardlyhel who pointed out the problem to me

## Problem

Liquid::Tokenizer and Liquid::C::Tokenizer differed in how they tokenized liquid tags.  It was parsing to a single token per-tag, including newlines.  However, Liquid::Tokenizer just used `@source.split("\n")` which omitted newlines and parsed to a single token per-line.

The newline character was preventing the `Liquid::BlockBody::LiquidTagToken` regex from matching, since `.` doesn't match the newline without it being a multiline regex, resulting in syntax errors when using the `disable_liquid_c_nodes: true` or `profile: true` parse option.

For example, parsing the following liquid tag with `disable_liquid_c_nodes`

```liquid
      {%- liquid
        assign x = 1
        assign y = x | plus: 2
        echo y
      -%}
```

would result in the following syntax error

```
Liquid::SyntaxError: Liquid syntax error: Unknown tag 'assign x = 1
  '
```

## Solution

Make Liquid::C::Tokenizer compatible with Liquid::Tokenizer by preserving the blank lines, but using a new token type for the C parsing code to easily ignore it.

### DIFF
--- a/test/unit/block_test.rb
+++ b/test/unit/block_test.rb
@@ -82,4 +82,26 @@ class BlockTest < MiniTest::Test
     end
     assert_equal(handler_error, exc)
   end
+
+  def test_liquid_tag
+    template = Liquid::Template.parse(<<~LIQUID)
+      {%- liquid
+        assign x = 1
+        assign y = x | plus: 2
+        echo y
+      -%}
+    LIQUID
+    assert_equal('3', template.render)
+  end
+
+  def test_liquid_tag_with_disable_liquid_c_nodes
+    template = Liquid::Template.parse(<<~LIQUID, disable_liquid_c_nodes: true)
+      {%- liquid
+        assign x = 1
+        assign y = x | plus: 2
+        echo y
+      -%}
+    LIQUID
+    assert_equal('3', template.render)
+  end
 end
--- a/test/unit/block_test.rb
+++ b/test/unit/block_test.rb
@@ -82,26 +82,4 @@ class BlockTest < MiniTest::Test
     end
     assert_equal(handler_error, exc)
   end
-
-  def test_liquid_tag
-    template = Liquid::Template.parse(<<~LIQUID)
-      {%- liquid
-        assign x = 1
-        assign y = x | plus: 2
-        echo y
-      -%}
-    LIQUID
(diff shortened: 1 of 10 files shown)


test/unit/block_test.rb
```diff
@@ -82,4 +82,26 @@ def test_exception_renderer_exception
     end
     assert_equal(handler_error, exc)
   end
+
+  def test_liquid_tag
```
macournoyer, line 86: Should this test be in liquid, and not liquid-c?

macournoyer (approved):
I think it will fix https://app.bugsnag.com/shopify/storefront-renderer/errors/5f9c18d6b00ff600186335a8?event_id=5f9c18d6006220d7ad810000&i=sk&m=nw that started happening recently in SFR!

peterzhu2118 (approved):

test/unit/block_test.rb
pushrax: It's a base case for the following test, which uses `disable_liquid_c_nodes: true`, so it's nice to have here right beside it.

dylanahsmith: There are already integration tests in the liquid repo that cover the same functionality.  This unit test just makes it clearer what is broken if the `test_liquid_tag_with_disable_liquid_c_nodes` test fails.

Actually, what would give us more confidence in `disable_liquid_c_nodes: true` would be to allow that to be set the default value for that option globally and to run all the integration tests with it defaulting to that value.  I tried doing that and it results in 5 LiquidTagTest test failures that this PR fixes.  I've just pushed a commit that does this instead of adding unit tests here.

pushrax (approved):
I didn't anticipate this combination of usage (tokenize with C + parse with Ruby), this fix makes sense to support that.

macournoyer (approved):

test/unit/block_test.rb
macournoyer: I didn't see a test case for `{% liquid ... assign`, but looking closer, other existing tests seem to cover :+1:

PR #119, opened 2020-11-04
Rename disable_liquid_c_nodes method to liquid_c_nodes_disabled?

dylanahsmith:
Since the former is a verb, which makes the call site look like it is disabling the liquid VM, rather than accessing an attribute.

### DIFF
--- a/lib/liquid/c.rb
+++ b/lib/liquid/c.rb
@@ -59,14 +59,14 @@ end
 
 Liquid::ParseContext.class_eval do
   class << self
-    attr_accessor :disable_liquid_c_nodes
+    attr_accessor :liquid_c_nodes_disabled
   end
-  self.disable_liquid_c_nodes = false
+  self.liquid_c_nodes_disabled = false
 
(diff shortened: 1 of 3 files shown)


lib/liquid/c.rb
```diff
@@ -59,30 +59,33 @@ def self.new(source, line_numbers = false, line_number: nil, for_liquid_tag: fal
...
+    if disabled_liquid_c_nodes
       ruby_new_block_body
     else
       Liquid::C::BlockBody.new
     end
   end
 
   # @api private
-  def disable_liquid_c_nodes
+  def disabled_liquid_c_nodes
```
peterzhu2118, line 77: IMO `disabled_liquid_c_nodes` sounds like it would contain a list of liquid-c nodes that are disabled. How about `liquid_c_nodes_disabled` and then we can do:

```suggestion
  def liquid_c_nodes_disabled?
```

peterzhu2118 (approved):

PR #120, opened 2020-11-05
Fix invalid float with no leading integer

peterzhu2118:
When fuzzing with AFL there was an inconsistency in parsing floats. Liquid does not allow floats without a leading integer part, but liquid-c will allow a negative float without a leading integer part (e.g. liquid-c will parse `-.1` as `-0.1`). I've added tests for this in liquid.

### DIFF
--- a/ext/liquid_c/lexer.c
+++ b/ext/liquid_c/lexer.c
@@ -109,7 +109,7 @@ const char *lex_one(const char *start, const char *end, lexer_token_t *token)
         }
     }
 
-    if (ISDIGIT(c) || c == '-') {
+    if (ISDIGIT(c) || (c == '-' && str + 1 < end && ISDIGIT(str[1]))) {
         int has_dot = 0;
         cur = str;
         while (++cur < end) {


ext/liquid_c/lexer.c
```diff
@@ -109,7 +109,7 @@ const char *lex_one(const char *start, const char *end, lexer_token_t *token)
         }
     }
 
-    if (ISDIGIT(c) || c == '-') {
+    if (ISDIGIT(c) || (c == '-' && str + 1 < end && ISDIGIT(str[1]))) {
```
dylanahsmith, line 112: This will make the strict parser more strict, so could be a breaking change.  We should instead use `Liquid::Usage.increment` to instrument this change to determine if it will break any code in practice.  If we determine that it is relied upon, then I think we will need to make the corresponding strict parsing in the ruby liquid gem less strict instead.

peterzhu2118: I opened #123 to instrument this usage.

PR #121, opened 2020-11-05
Fix negative string length

peterzhu2118:
AFL caught a case that causes liquid-c to attempt to create a Ruby string with negative length.

When you try to parse liquid `{%-%}`, you get a

```
ArgumentError: negative string size (or size too big)
```

This is because the tokenizer treats `{%-%}` as if it has content inside of it, which it does not.

### DIFF
--- a/ext/liquid_c/tokenizer.c
+++ b/ext/liquid_c/tokenizer.c
@@ -1,3 +1,4 @@
+#include <assert.h>
 #include "liquid.h"
 #include "tokenizer.h"
 #include "stringutil.h"
@@ -189,8 +190,12 @@ found:
 
     if (token->type == TOKEN_VARIABLE || token->type == TOKEN_TAG) {
         token->str_trimmed += 2 + token->lstrip;
-        token->len_trimmed -= 2 + token->lstrip + 2 + token->rstrip;
+        token->len_trimmed -= 2 + token->lstrip + 2;
+        if (token->rstrip && token->len_trimmed)
+            token->len_trimmed--;
     }
+
+    assert(token->len_trimmed >= 0);
 
     tokenizer->cursor += token->len_full;
 
--- a/test/unit/tokenizer_test.rb
+++ b/test/unit/tokenizer_test.rb
@@ -43,6 +43,11 @@ class TokenizerTest < Minitest::Test
     assert_equal(["", "funk", "", "so | brother"], tokenize(source, for_liquid_tag: true, trimmed: true))
   end
 
+  def test_invalid_tags
+    assert_equal([""], tokenize("{%-%}", trimmed: true))
+    assert_equal([""], tokenize("{{-}}", trimmed: true))
+  end
+
   def test_utf8_encoded_source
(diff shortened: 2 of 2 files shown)


test/unit/block_test.rb
```diff
@@ -31,6 +31,10 @@ def test_op_write_raw_w
     assert_equal(source, template.render!)
   end
 
+  def test_invalid_blocks
```
dylanahsmith, line 34: It seems like this is testing an invalid tag.  Also, it seems like the test belongs in test/unit/tokenizer_test.rb, since the code being tested is in the tokenizer.

test/unit/block_test.rb
```diff
@@ -31,6 +31,10 @@ def test_op_write_raw_w
...
+    assert_raises(Liquid::SyntaxError) { Liquid::Template.parse("{%-%}") }
```
dylanahsmith, line 35: We should have a similar test for an empty variable with a single strip character (`Liquid::Template.parse("{{-}}")`) which I don't think results in a syntax error.  That also brings implications of how stripping should be handled, since this seems like an edge case that could result in stripping only one side or both sides.

ext/liquid_c/tokenizer.c
```diff
@@ -141,16 +141,20 @@ static void tokenizer_next_for_template(tokenizer_t *tokenizer, token_t *token)
...
                 while (c == '%' && cursor <= last)
                     c = *cursor++;
                 if (c != '}')
                     continue;
                 token->type = TOKEN_TAG;
-                if(cursor[-3] == '-')
+                if (has_body && cursor[-3] == '-')
                     token->rstrip = tokenizer->lstrip_flag = true;
                 goto found;
             }
```
dylanahsmith, line 160: We also need to fix the code path for variables.  If you add `assert(token->len_trimmed > 0);` after it has been computed at the end of the function, then that will catch the problem for variables with a `Liquid::Template.parse("{{-}}")` test.

dylanahsmith (changes requested):

ext/liquid_c/tokenizer.c
```diff
@@ -189,9 +195,11 @@ static void tokenizer_next_for_template(tokenizer_t *tokenizer, token_t *token)
 
     if (token->type == TOKEN_VARIABLE || token->type == TOKEN_TAG) {
         token->str_trimmed += 2 + token->lstrip;
-        token->len_trimmed -= 2 + token->lstrip + 2 + token->rstrip;
+        token->len_trimmed -= 4 + (has_body ? token->lstrip + token->rstrip : (token->lstrip | token->rstrip));
```
dylanahsmith, line 198: We don't need `has_body`, we just need to make sure we aren't decrementing len_trimmed below 0 here.

```suggestion
        token->len_trimmed -= 2 + token->lstrip + 2;
        if (token->rstrip && token->len_trimmed)
          token->len_trimmed--;
```

dylanahsmith (approved):

PR #122, opened 2020-11-05
Allow space in between dot in attributes

peterzhu2118:
Liquid allows any number of spaces between the dot when referring to an attribute but liquid-c raises a syntax error. I've included a test in liquid.

### DIFF
--- a/ext/liquid_c/parser.c
+++ b/ext/liquid_c/parser.c
@@ -128,11 +128,8 @@ static void parse_and_compile_variable_lookup(parser_t *p, vm_assembler_t *code)
             parser_must_consume(p, TOKEN_CLOSE_SQUARE);
             vm_assembler_add_lookup_key(code);
         } else if (p->cur.type == TOKEN_DOT) {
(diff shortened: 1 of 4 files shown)


ext/liquid_c/parser.c
```diff
@@ -128,12 +128,9 @@ static void parse_and_compile_variable_lookup(parser_t *p, vm_assembler_t *code)
             parser_must_consume(p, TOKEN_CLOSE_SQUARE);
             vm_assembler_add_lookup_key(code);
         } else if (p->cur.type == TOKEN_DOT) {
-            int has_space_affix = parser_consume_any(p).flags & TOKEN_SPACE_AFFIX;
```
dylanahsmith, line 131: I think this makes TOKEN_SPACE_AFFIX, TOKEN_SPACE_PREFIX and TOKEN_SPACE_SUFFIX unused, so I think we can remove them now.

dylanahsmith (approved):
Other than unused code cleanup, LGTM

dylanahsmith (approved):

PR #123, opened 2020-11-06
Instrument float bug

peterzhu2118:
Instruments bug #120 to determine the usage of it in production.


dylanahsmith (approved):

tobi:
what did we learn? Can we merge the bugfix?

peterzhu2118:
It looks like there are still stores using this feature. I think we should fix these store/let them know about this breaking change before we fix the bug.

Screen Shot 2022-01-01 at 3 05 42 PM

PR #124, opened 2020-11-10
Remove error on compiler warnings by default, except on CI

dylanahsmith:
## Problem

@tobi encountered a compiler warning that was treated as an error due when trying to compile liquid-c on linux with gcc 9.3.

This wasn't an error that was caught by CI and I wasn't able to reproduce this and I don't think we should have confidence that we can catch all compiler warnings before another developer experiences them.

## Solution

Only treat warnings as errors by default on CI.  We can set the `LIQUID_C_PENDANTIC` environment variable to opt-in to this locally.


peterzhu2118 (approved):

dylanahsmith:
Tobi already committed the fix for this compiler error in commit b4353415da934da28d6f13df2728f1fd79e00427

PR #128, opened 2020-11-25
Rename per-file extension init methods to distinguish from object init

dylanahsmith:
## Problem

We have two sets of initialization functions.  We have object initialization functions and functions to MRI extension initialization functions.  These used to be subtle distinguished by one using an init_liquid_ prefix and one using an _init suffix, however, `vm_assembler_pool_init` was introduced which doesn't even follow this convention.

## Solution

The per-file MRI extension initialization functions are mostly defining classes, modules, methods, etc.  So use a `define_` prefix for these after the `liquid_` namespace prefix (meant to avoid naming conflicts).

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -531,7 +531,7 @@ static VALUE block_body_add_filter(VALUE self, VALUE filter_name, VALUE num_args
 }
 
 
-void init_liquid_block()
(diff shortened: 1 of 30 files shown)


peterzhu2118 (approved):
The inconsistency in `vm_assembler_pool_init` was my bad. Sorry about that!

PR #134, opened 2020-12-14
Implement raw tag parsing in BlockBody

peterzhu2118:
See Shopify/liquid#1378.

# Benchmarks

In a microbenchmark for `raw` tags, this PR introduces a performance impact during parsing.

This branch:

```
               parse     80.202k (± 8.7%) i/s -    400.816k in   5.036587s
              render    214.678k (± 6.1%) i/s -      1.081M in   5.055158s
      parse & render     48.202k (± 5.8%) i/s -    243.744k in   5.074713s
```

Master:

```
               parse    105.637k (± 6.9%) i/s -    535.392k in   5.092843s
              render    206.248k (± 8.5%) i/s -      1.036M in   5.058648s
      parse & render     57.954k (± 6.6%) i/s -    292.464k in   5.069258s
```

Script:

```ruby
require "benchmark/ips"
require "liquid"
require "liquid/c"

source = <<~LIQUID
  {% raw %}
    {% test %}
    hello world
    {% endtest %}
  {% endraw %}
LIQUID

template = Liquid::Template.parse(source)

Benchmark.ips do |x|
  x.report("parse") do |times|
    times.times do
      Liquid::Template.parse(source)
    end
  end

  x.report("render") do |times|
    times.times do
      template.render!
    end
  end

  x.report("parse & render") do |times|
    times.times do
      Liquid::Template.parse(source).render!
    end
  end
end
```


ext/liquid_c/block.c
```diff
@@ -125,6 +125,90 @@ static int is_id(int c)
...
+static bool match_full_token_possibly_invalid(token_t *token, full_token_possibly_invalid_t *match)
+{
+    const char *str = token->str_full;
+    long len = token->len_full;
+
+    match->body_len = 0;
+    match->delimiter_start = NULL;
+    match->delimiter_len = 0;
+
+    if (len < 5) return false; // Must be at least 5 characters: \{%\w%\}
```
macournoyer, line 143: Although it is only called for raw tags at the moment, the name implies it could be used for parsing other tags. So, could it be called within `{% liquid ... %}`? In which case the tags would not be wrapped in `{% ... %}`.

macournoyer (approved):
Looks like parsing code was simply moved to block.c 👍 

I think parsing performance will soon be irrelevant with serialization 🎉

peterzhu2118:
Closing since #135 is merged.

PR #135, opened 2020-12-16
Write raw tags directly to buffer

peterzhu2118:
Write raw tags directly to the instructions buffer as an `OP_WRITE_RAW` instruction. This is an alternative to #134.

# Benchmarks

There is a small performance hit during parsing, but a slight performance improvement during rendering in the microbenchmark.

Master:

```
               parse    105.637k (± 6.9%) i/s -    535.392k in   5.092843s
              render    206.248k (± 8.5%) i/s -      1.036M in   5.058648s
      parse & render     57.954k (± 6.6%) i/s -    292.464k in   5.069258s
```

This branch:

```
               parse     97.280k (± 3.3%) i/s -    486.423k in   5.006012s
              render    225.835k (± 6.2%) i/s -      1.126M in   5.009018s
      parse & render     53.268k (± 8.9%) i/s -    268.041k in   5.077031s
```

Script:

```ruby
require "benchmark/ips"
require "liquid"
require "liquid/c"

source = <<~LIQUID
  {% raw %}
    {% test %}
    hello world
    {% endtest %}
  {% endraw %}
LIQUID

template = Liquid::Template.parse(source)

Benchmark.ips do |x|
  x.report("parse") do |times|
    times.times do
      Liquid::Template.parse(source)
    end
  end

  x.report("render") do |times|
    times.times do
      template.render!
    end
  end

  x.report("parse & render") do |times|
    times.times do
      Liquid::Template.parse(source).render!
    end
  end
end
```

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -248,7 +248,17 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
                 if (body->as.intermediate.blank && !RTEST(rb_funcall(new_tag, intern_is_blank, 0)))
                     body->as.intermediate.blank = false;
 
-                vm_assembler_add_write_node(body->as.intermediate.code, new_tag);
+                if (tokenizer->raw_tag_body) {
+                    if (tokenizer->raw_tag_body_len) {
+                        vm_assembler_add_write_raw(body->as.intermediate.code, tokenizer->raw_tag_body,
+                                                tokenizer->raw_tag_body_len);
+                    }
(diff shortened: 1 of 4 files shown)


macournoyer (approved):
I think this one is better then 134 & https://github.com/Shopify/liquid/pull/1378 on all fronts. It's simpler, faster, and since the goal it to move to byte-code, I don't see why we would do it any other way?

dylanahsmith (approved):
Makes sense as an approach before https://github.com/Shopify/liquid-c/pull/96/ is merged

PR #137, opened 2020-12-16
Implement serialize

peterzhu2118:
Based on #136.

This PR implements `BlockBody#dump` that dumps the block body to a string. The block body dumps the instructions, writes the tag markups to the buffer, and marshal dumps the constants into a single string that is returned.


ext/liquid_c/document_body.h
```diff
@@ -28,6 +28,14 @@ typedef struct document_body {
     c_buffer_t buffer;
 } document_body_t;
 
+typedef struct document_body_header {
+    uint32_t entrypoint_block_index;
```
macournoyer, line 32: I see it's called index here, but it ends up being used as an offset here: https://github.com/Shopify/liquid-c/pull/138/files#diff-d25b9c9a1b16f1935c261f8232ea43e790513f022f33b7c8de117016cd76f130R188 correct?

Should this be renamed to _offset or the code is wrong in #138?

macournoyer (approved):

ext/liquid_c/document_body.c
```diff
@@ -129,6 +129,34 @@ void document_body_write_block_body(VALUE self, bool blank, uint32_t render_scor
...
+
+    uint32_t buffer_len = (uint32_t)c_buffer_size(&body->buffer);
+
+    VALUE constants = rb_marshal_dump(body->constants, Qnil);
+    uint32_t constants_len = (uint32_t)RSTRING_LEN(constants);
+
+    VALUE str = rb_str_buf_new(sizeof(document_body_header_t) + buffer_len + constants_len);
+
+    document_body_header_t header = {
+        .entrypoint_block_index = entrypoint_block_index,
```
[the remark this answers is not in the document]
dylanahsmith: Yes.  I suggested this in the deserialize PR here https://github.com/Shopify/liquid-c/pull/138/files#r551963712

ext/liquid_c/document_body.h
dylanahsmith: You're right, it is an offset.  There isn't currently a block table to index into.

PR #138, opened 2020-12-16
Implement deserialize

peterzhu2118:
Based on #137.

This PR implements `Liquid::Template.load` which reads and deserializes a parse liquid template.

I added a `SerializeParseContext` class that is a subclass of `ParseContext` that is used to store the information required for deserialization. The actual deserialization work is done in the function `block_body_parse_from_serialize` that is called during `Liquid::BlockBody.parse` when a `SerializeParseContext` is passed in.


test/unit/template_test.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class TemplateTest < MiniTest::Test
+  def test_serialize
```
peterzhu2118, line 6: I'm not sure how to (or need to) further improve test coverage.

tobi:
Great work Peter.

peterzhu2118:
Thank you Tobi! ♥

test/unit/template_test.rb
macournoyer: I think if we add this to the benchmarks, and run those as part of CI it would add coverage for free.

We also need to make sure `Liquid::Template.load(serialize).render!` is, and will remain faster than `Liquid::Template.parse(source).render!`.

peterzhu2118: I implemented benchmarks in Shopify/liquid#1385. Deserialize is about 20% faster than parsing from source.

ext/liquid_c/block.c
```diff
@@ -290,32 +318,87 @@ static VALUE internal_block_body_parse(block_body_t *body, parse_context_t *pars
...
 }
 
-static void ensure_intermediate(block_body_t *body)
+typedef struct block_body_yield_tag_args {
+    block_body_t *body;
+    serialize_parse_context_t *serialize_context;
+    tag_markup_header_t *current_tag;
+} block_body_yield_tag_args_t;
+
+static VALUE block_body_try_yield_tag(VALUE uncast_args)
```
macournoyer, line 327: Still make sense to have `try` in the name since the `raise` was removed?

ext/liquid_c/block.c
```diff
@@ -290,32 +318,87 @@ static VALUE internal_block_body_parse(block_body_t *body, parse_context_t *pars
...
+                .serialize_context = serialize_context,
+                .current_tag = current_tag
+            };
+            return rb_rescue(block_body_try_yield_tag, (VALUE)&yield_args, block_body_rescue_yield_tag, (VALUE)&yield_args);
+        } else {
+            VALUE tag_name = rb_utf8_str_new(tag_markup_header_name(current_tag), current_tag->tag_name_len);
+            VALUE markup = rb_utf8_str_new(tag_markup_header_markup(current_tag), current_tag->markup_len);
+
+            VALUE tag_class = rb_funcall(tag_registry, intern_square_brackets, 1, tag_name);
+            assert(RTEST(tag_class));
```
macournoyer, line 378: I think this should raise? Seems technically possible that tag could get removed between the time the template is parsed, cached (for a long time) & deserialized.

Maybe using `Block.unknown_tag`?

macournoyer (approved):
Wow! That was a big piece. Nice work 🙇 

The only potential issue I see is w/ the `assert(RTEST(tag_class));`

ext/liquid_c/block.c
peterzhu2118: I think it does make sense to have `try` in the name since it's called in the context of a `try/rescue` [here](https://github.com/Shopify/liquid-c/blob/ed940df097a2a8d55c7915c726a3242496267c20/ext/liquid_c/block.c#L372).

ext/liquid_c/block.c
peterzhu2118: :+1: Good point. On a unrelated note, should `block.c` be renamed to `block_body.c` since it actually contains the code for `Liquid::C::BlockBody`?

ext/liquid_c/tag_markup.c
```diff
@@ -65,6 +65,36 @@ void tag_markup_set_block_body(VALUE self, VALUE block_body_obj, block_body_t *b
     tag->block_body = block_body;
 }
 
+tag_markup_header_t *tag_markup_get_next_tag(document_body_entry_t *entry, tag_markup_header_t *current_tag)
```
dylanahsmith, line 68: It doesn't seem like we should even need the entry pointer to get the next tag, since the next tag is immediately after the current one in memory.  Also, it seems like this function would be simpler if it didn't have to handle getting the first tag.

Suggested change: https://github.com/Shopify/liquid-c/commit/a46feef01ca7183ff293bc997c694042d1636375

ext/liquid_c/block.c
dylanahsmith: yes, block_body.c would be a more appropriate name for this file.

ext/liquid_c/serialize_parse_context.c
```diff
@@ -0,0 +1,76 @@
...
+    obj = TypedData_Make_Struct(cLiquidCSerializeParseContext, serialize_parse_context_t,
+                                &serialize_parse_context_data_type, serialize_context);
+    assert(header->entrypoint_block_index < header->buffer_len);
+    serialize_context->document_body = document_body;
+    document_body_setup_entry_for_header(document_body, header->entrypoint_block_index,
+                                         &serialize_context->current_entry);
+    serialize_context->current_tag = tag_markup_get_next_tag(&serialize_context->current_entry, NULL);
+
+    // Call initialize method of parent class
+    rb_funcall(obj, id_initialize, 0);
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in 54678bda70dce146cd3984a93195aa36a039b3de

ext/liquid_c/template.c
```diff
@@ -0,0 +1,52 @@
...
+{
+    VALUE str_obj = rb_str_new_static(str, len);
+    VALUE constants = rb_marshal_load(str_obj);
+    if (BUILTIN_TYPE(constants) != T_ARRAY) {
+        rb_raise(rb_eArgError, "expected constants to be an array");
+    }
+    return constants;
+}
+
+static VALUE template_load(VALUE self, VALUE source, VALUE options)
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in 6dc5d6b25e82af5d21be99b5b219386189791672

ext/liquid_c/template.c
```diff
@@ -0,0 +1,52 @@
...
+{
+    rb_funcall(self, id_configure_options, 1, options);
+
+    Check_Type(source, T_STRING);
+    source = rb_str_dup_frozen(source);
+    const char *data = RSTRING_PTR(source);
+
+    document_body_header_t *header = (document_body_header_t *)data;
+
+    assert(RSTRING_LEN(source) >= header->buffer_offset + header->buffer_offset);
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in a406547c7108626167564349945ba5e2a9876b2b

test/unit/template_test.rb
```diff
@@ -0,0 +1,22 @@
...
+    assert_equal('hello world', dump_load_eval('hello world'))
+    assert_equal('hello world', dump_load_eval('{% assign greeting = "hello" %}{{ greeting }} world'))
+    assert_equal('hello world', dump_load_eval('{% raw %}hello {% endraw %}world'))
+    assert_equal('hello world',
+      dump_load_eval('{% if test %}goodbye {% else %}hello {% endif %}world', 'test' => false))
+    assert_equal('hello world', dump_load_eval('{% if true %}hello {% endif %}{% if true %}world{% endif %}'))
+    assert_equal('123', dump_load_eval('{% for i in (1..10) %}{{i}}{% if i == 3 %}{% break %}{% endif %}{% endfor %}'))
+  end
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in 005b088cec30364d5f322e4d45d161008cd045e7. I added a `Liquid::C::DeserializationError` class that is raised for all serialization incompatibility reasons.

ext/liquid_c/template.c
```diff
@@ -0,0 +1,52 @@
...
+static VALUE template_load(VALUE self, VALUE source, VALUE options)
+{
+    rb_funcall(self, id_configure_options, 1, options);
+
+    Check_Type(source, T_STRING);
+    source = rb_str_dup_frozen(source);
+    const char *data = RSTRING_PTR(source);
+
+    document_body_header_t *header = (document_body_header_t *)data;
+
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in 005b088cec30364d5f322e4d45d161008cd045e7. I added a `Liquid::C::DeserializationError` class that is raised for all serialization incompatibility reasons.

ext/liquid_c/block.c
```diff
@@ -290,32 +318,87 @@ static VALUE internal_block_body_parse(block_body_t *body, parse_context_t *pars
...
+    while (current_tag) {
+        bool tag_unknown = TAG_UNKNOWN_P(current_tag);
+
+        if (tag_unknown) {
+            block_body_yield_tag_args_t yield_args = {
+                .body = body,
+                .serialize_context = serialize_context,
+                .current_tag = current_tag
+            };
+            return rb_rescue(block_body_try_yield_tag, (VALUE)&yield_args, block_body_rescue_yield_tag, (VALUE)&yield_args);
```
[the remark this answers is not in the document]
peterzhu2118: Fixed in bffc52c4b602e2d6d05a5aae11c2a335c90834a2

ext/liquid_c/tag_markup.c
peterzhu2118: Fixed in deb3c0acb98a0c8ecf82d1086ad61dfc2ba251ee

test/unit/template_test.rb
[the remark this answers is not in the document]
peterzhu2118: Is e3f403c6943d398406a9b3d974818e473362ced7 the kind of test you were thinking of?

peterzhu2118:
@dylanahsmith I added a rake task that runs the liquid integration tests for serialization in a65f10e85ea7d73da23c36bb68c951e2f4ddadba and it's now running as part of CI. The commits after that commit are fixes I made from bugs discovered when running the integration tests.

PR #140, opened 2021-01-07
Make Liquid::C::Tokenizer#shift private so it is only used for testing

dylanahsmith:
Depends on https://github.com/Shopify/liquid/pull/1386

## Problem

The liquid template (de)serialization support that is being added by https://github.com/Shopify/liquid-c/pull/138 won't actually support `Liquid::C::Tokenizer#shift` being used during parsing.  The method would seem to work for the initial parsing, but then it would fail on deserialization where there is no longer a tokenizer.

## Solution

Leverage https://github.com/Shopify/liquid/pull/1386 to update the tokenizer monkey patch to instead be on Liquid::ParseContext.new_tokenizer so that we avoid using the Liquid::C::Tokenizer when the liquid-c VM is disabled.

Make Liquid::C::Tokenizer#shift private, like `shift_trimmed`, and only use it for testing.

### DIFF
--- a/lib/liquid/c.rb
+++ b/lib/liquid/c.rb
@@ -41,20 +41,16 @@ module Liquid
   end
 end
 
-Liquid::Tokenizer.class_eval do
-  def self.new(source, line_numbers = false, line_number: nil, for_liquid_tag: false)
-    source = source.to_s
-    if Liquid::C.enabled && source.bytesize <= Liquid::C::Tokenizer::MAX_SOURCE_BYTE_SIZE
-      source = source.encode(Encoding::UTF_8)
-      Liquid::C::Tokenizer.new(source, line_number || (line_numbers ? 1 : 0), for_liquid_tag)
-    else
-      super
-    end
-  end
-end
-
 Liquid::Raw.class_eval do
   alias_method :ruby_parse, :parse
+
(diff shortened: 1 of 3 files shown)


peterzhu2118 (approved):

lib/liquid/c.rb
```diff
@@ -73,34 +69,42 @@ def new_block_body
...
   end
 
+  alias_method :ruby_new_tokenizer, :new_tokenizer
+
+  def new_tokenizer(source, line_number: nil, for_liquid_tag: false)
+    unless liquid_c_nodes_disabled?
+      source = source.to_s
+      if source.bytesize <= Liquid::C::Tokenizer::MAX_SOURCE_BYTE_SIZE
+        source = source.encode(Encoding::UTF_8)
+        return Liquid::C::Tokenizer.new(source, line_number || 0, for_liquid_tag)
```
macournoyer, line 79: Shouldn't it be `|| 1`?

macournoyer (approved):

lib/liquid/c.rb
dylanahsmith: The C code uses `0` to mean that line numbers aren't tracked, similar to if `line_number` is `nil` in the ruby tokenizer.

PR #144, opened 2021-02-19
Add support for inline comments

dylanahsmith:
## Problem

Adding a `#` inline comment tag for https://github.com/Shopify/liquid/issues/1393 requires changing the parsing code to recognize that as a tag.

## Solution

Change the code path for where no word characters are found for tag name, so this shouldn't even impact the performance of handling non-comment tags.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -199,9 +199,14 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
                 long name_len = name_end - name_start;
 
                 if (name_len == 0) {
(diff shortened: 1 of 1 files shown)


pushrax (approved):

samdoiron:
I'm seeing a test failure in Liquid here when combined with [this branch](https://github.com/Shopify/liquid/pull/1498)

https://github.com/Shopify/liquid/blob/05dc7d7de439415de50f3442c7bc0979affa6a29/test/integration/variable_test.rb#L20

This looks unrelated, though. Perhaps a bug on main?

samdoiron:
This may also be because the branches are out of date, checking that now.

PR #146, opened 2021-03-11
Provide test/liquid_test_helper to make it easier to run specific liquid tests

dylanahsmith:
## Problem

When there are integration test failures, it can be a bit awkward to run a specific test in the liquid test suite.  It isn't as simple as using `bundle exec ruby -Itest`, since the load path must be setup with both the liquid repo's test path and the liquid-c repo's lib path.

## Solution

Extract the setup for test/integration_test.rb into a test/liquid_test_helper file so that it can be required to do this setup.  It can be used to run a specific test using `ruby -r./test/liquid_test_helper` followed by the test file and any minitest arguments.

For example:
```
ruby -r./test/liquid_test_helper `bundle info liquid --path`/test/integration/template_test.rb
```

### DIFF
--- a/test/liquid_test_helper.rb
+++ b/test/liquid_test_helper.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# This can be used to setup for running tests from the liquid test suite.
+# For example, you could run a single liquid test as follows:
+# $ ruby -r./test/liquid_test_helper `bundle info liquid --path`/test/integration/template_test.rb
+
+require 'bundler/setup'
+
+liquid_lib_dir = $LOAD_PATH.detect { |p| File.exist?(File.join(p, 'liquid.rb')) }
+LIQUID_TEST_DIR = File.join(File.dirname(liquid_lib_dir), 'test')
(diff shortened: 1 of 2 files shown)


test/liquid_test_helper.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# This can be used to setup for running tests from the liquid test suite.
+# For example, you could run a single liquid test as follows:
+# $ ruby -r./test/liquid_test_helper `bundle info liquid --path`/test/integration/template_test.rb
```
macournoyer, line 5: TIL! I guess this can't be a rake task or such in this repo because it needs to run the bundler context of Liquid?

macournoyer (approved):

peterzhu2118 (approved):

test/liquid_test_helper.rb
dylanahsmith: There is already a rake task for running the liquid integration tests, however, that is for running all integration tests rather than specific ones.

PR #147, opened 2021-03-11
Return document body entry from document_body_write_block_body

dylanahsmith:
Depends on https://github.com/Shopify/liquid-c/pull/145

In https://github.com/Shopify/liquid-c/pull/145, I was considering adding a `rb_check_frozen(document_body);` check into `document_body_write_block_body` but realized an exception in the function would leave the block body in an bad state, since it was in the middle of changing the union from the intermediate to the compiled state.

In order to avoid this type of problem, which is theoretically also possible on out of memory exceptions, we can instead return the document_body_entry_t so that the block body struct can have its union changed all at once after document_body_write_block_body returns.

### DIFF
--- a/ext/liquid_c/block.c
+++ b/ext/liquid_c/block.c
@@ -317,9 +317,9 @@ static VALUE block_body_freeze(VALUE self)
     bool blank = body->as.intermediate.blank;
     uint32_t render_score = body->as.intermediate.render_score;
     vm_assembler_t *code = body->as.intermediate.code;
-    body->compiled = true;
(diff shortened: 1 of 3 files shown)


peterzhu2118 (approved):

dylanahsmith:
> which is theoretically also possible on out of memory exceptions

Looks like this was more than just a theoretical problem, since I was able to cause crashes without this change by putting `rb_gc_start();` in the `document_body_write_block_body` function at any of the allocation points, where allocation can trigger garbage collection.  I'll cherry-pick this into the 2020-stable branch

PR #152, opened 2021-05-20
use to_liquid_value when we are accessing a hash or an array with a Drop

ggmichaelgo:
This is a C implementation for this change in Liquid PR:
https://github.com/Shopify/liquid/pull/1441


ext/liquid_c/variable_lookup.c
```diff
@@ -1,10 +1,14 @@
...
 #include "context.h"
 
-static ID id_has_key, id_aref, id_fetch;
+static ID id_has_key, id_aref, id_fetch, id_to_liquid_value;
 
 VALUE variable_lookup_key(VALUE context, VALUE object, VALUE key, bool is_command)
 {
+    if (key && rb_respond_to(key, id_to_liquid_value)) {
+        key = rb_funcall(key, id_to_liquid_value, 0);
+    }
```
dylanahsmith, line 10: I don't see the need to check if C is non-zero.  It should be usable with rb_respond_to no matter what ruby value it is.

Also, you can use `rb_check_funcall` to call a method if it exists, which should avoid having to have the method get looked up twice to call it.

```suggestion
    VALUE key_value = rb_check_funcall(key, id_to_liquid_value, 0);
    if (key_value != Qundef)
        key = key_value;
```

test/unit/variable_test.rb
```diff
@@ -232,6 +232,59 @@ def test_filter_argument_error_translation
...
+    def ==(other)
+      other == @value
+    end
+
+    def to_s
+      @value ? 'Yay' : 'Nay'
+    end
+  end
+
+  def test_to_liquid_value_on_variable_lookup_and_if_statement
```
dylanahsmith, line 275: This isn't liquid-c specific, so it belongs in the liquid repo under the test/integration/ directory and it will get tested with liquid-c

ext/liquid_c/variable_lookup.c
dylanahsmith: I think the common case is when the key is a string.  We may want to leverage that fast path only trying to call the method if `rb_obj_class(key) != rb_cString`.

A more important optimization would be to avoid this check entirely if it is a constant key, since there is already special cases for that for the `OP_LOOKUP_CONST_KEY` and `OP_LOOKUP_COMMAND` VM instructions.

PR #159, opened 2021-10-14
Do not use liquid-c when using fallback parsing

peterzhu2118:
Using a liquid (Ruby, non-c) tokenizer falls back to liquid to parse, we should set liquid_c_nodes_disabled on the ParseContext to ensure that nothing is parsed through liquid-c.


peterzhu2118:
Closing because #160 has merged.

PR #174, opened 2022-02-28
Franklyn/reviewable

damnMeddlingKid:


ext/liquid_c/block.c
```diff
@@ -262,6 +282,112 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
...
+          - resolve the address for the previous OP_BRANCH_UNLESS
+        6 on endif resolve the address for any OP_BRANCH/OP_BRANCH_UNLESS
+    */
+    vm_assembler_t* body_code = body->as.intermediate.code;
+    VALUE condition_obj = parse_single_binary_comparison(markup);
+    vm_assembler_add_op_with_constant(body_code, condition_obj, OP_EVAL_CONDITION);
+
+    uint8_t* instruction;
+
+    ptrdiff_t exit_branches[10];
```
damnMeddlingKid, line 323: need to use a dynamically sized list here, something like cbuffer.

ext/liquid_c/block.c
```diff
@@ -262,6 +282,112 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
     return unknown_tag;
 }
 
+VALUE parse_single_binary_comparison(VALUE markup) {
```
damnMeddlingKid, line 285: Only supports parsing 1 binary comparison, need to add more parsing code to support full conditions.

ext/liquid_c/block.c
```diff
@@ -262,6 +282,112 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
...
+            StringValue(unknown_tag.name);
+            char *name_start = RSTRING_PTR(unknown_tag.name);
+            int name_len = RSTRING_LEN(unknown_tag.name);
+
+            if (name_len == 4 && strncmp(name_start, "else", 4) == 0) {
+                // Unconditionally branch to endif for the previous block
+                *exit_end++ = vm_assembler_open_branch(body_code, OP_BRANCH);
+
+                // Calculate the offset that would jump to here, this is where the <if> jumps to if it fails the condition.
+                jump = (ptrdiff_t) (body_code->instructions.data_end - body_code->instructions.data);
```
damnMeddlingKid, line 345: Calculating indexes since cbuffer can realloc.

ext/liquid_c/block.c
```diff
@@ -262,6 +282,112 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
...
+    if(op.type) {
+        VALUE op_str = rb_enc_str_new(op.val, op.val_end - op.val, utf8_encoding);
+        VALUE b = internal_expression_parse(&p);
+        return rb_funcall(cLiquidCondition, intern_new, 3, a, op_str, b);
+    } 
+
+    return rb_funcall(cLiquidCondition, intern_new, 1, a);
+}
+
+tag_markup_t parse_if_tag(VALUE markup, block_body_t *body, parse_context_t *parse_context) {
```
damnMeddlingKid, line 306: theres no validation yet for guarding against badly structured if, for example an elsif without if.

ext/liquid_c/block.c
damnMeddlingKid: The offsets get stored with the instruction in 16 bits. assuming this is ok for a jump offset, otherwise need to store them somewhere else.

ext/liquid_c/block.c
```diff
@@ -222,11 +219,34 @@ static tag_markup_t internal_block_body_parse(block_body_t *body, parse_context_
...
+
+                if (name_len == 2 && strncmp(name_start, "if", 2) == 0) {
+                    unknown_tag = parse_if_tag(markup, body, parse_context);
+                    if (unknown_tag.name != Qnil) {
+                        goto loop_break;
+                    }
+                    render_score_increment += 1;
+                    body->as.intermediate.blank = false;
+                    break;
+                } else if (
```
damnMeddlingKid, line 240: handling unknown tags here could get messy as we introduce other tags, need a better abstraction for this.

damnMeddlingKid: I was thinking of something along the lines of https://github.com/Shopify/liquid-c/pull/96 so we could pick the right parsing handler from ruby land.

PR #178, opened 2022-07-18
Remove liquid expression parse patch

dylanahsmith:
## Problem

@samdoiron found a bug in https://github.com/Shopify/liquid-c/pull/154, where `parse_context.parse_expression` would not respect `disable_liquid_c_nodes: true`, because even though that would make sure that Liquid::Expression.ruby_parse gets used, it would internally call the hard coded `Liquid::Expression.parse` that would use `Liquid::C::Expression.lax_parse`.

## Solution

Get rid of the Liquid::Expression.parse monkey patch and instead require the application to use `Liquid::C::Expression.parse` if they really want to create a constant Liquid::C::Expression object.

### DIFF
--- a/test/unit/expression_test.rb
+++ b/test/unit/expression_test.rb
@@ -159,7 +159,7 @@ class ExpressionTest < MiniTest::Test
   end
 
   def test_disable_c_nodes
-    context = Liquid::Context.new({ "x" => 123 })
+    context = Liquid::Context.new({ "x" => 123, "y" => { 123 => 42 } })
 
     expr = Liquid::ParseContext.new.parse_expression("x")
     assert_instance_of(Liquid::C::Expression, expr)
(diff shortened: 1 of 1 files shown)


.github/workflows/liquid.yml
```diff
@@ -6,7 +6,6 @@ jobs:
     strategy:
       matrix:
         entry:
-          - { ruby: '2.5', allowed-failure: false }
```
dylanahsmith, line 9: This is needed because of ruby liquid dropping support for EOL ruby versions (https://github.com/Shopify/liquid/pull/1578)

samdoiron:
I'm not sure this solved the bug. When I try with a clone locally:

```ruby
irb> Liquid::Template.parse("{{ foo[bar] }}", disable_liquid_c_nodes: true)
  .root.body.nodelist[0].name.lookups[0].class
=> Liquid::C::Expression
```

PR #181, opened 2022-09-29
Add support for parse-specific tags

dylanahsmith:
liquid-c support for https://github.com/Shopify/liquid/pull/1629

The only difference is that it gets the tags from Liquid::ParseContext#tags instead of from Liquid::Tempate.tags.


dylanahsmith:
Closing since #1629 was closed

PR #190, opened 2023-03-29
Fix crash in the VM from garbage collected object

peterzhu2118:
vm_stack_pop_n_use_in_place removes objects from the VM stack. If these objects do not have any other references, then they can be garbage collected, leading to T_NONE objects being passed into vm_invoke_filter.

This commit fixes this issue in OP_HASH_NEW and OP_FILTER by first peeking from the VM stack and only pop after it is safe to do so.

Here's a stack trace of a crash:

    (gdb) bt 20
    #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
    #1  0x00007f0efd742859 in __GI_abort () at abort.c:79
    #2  0x000055cf5fdc5d4f in die () at error.c:817
    #3  rb_bug_for_fatal_signal (default_sighandler=0x0, sig=sig@entry=11, ctx=ctx@entry=0x7f0efc743900, fmt=fmt@entry=0x55cf601e37aa "Segmentation fault at %p") at error.c:817
    #4  0x000055cf5ff2872d in sigsegv (sig=11, info=0x7f0efc743a30, ctx=0x7f0efc743900) at signal.c:964
    #5  0x00007f0efd941420 in <signal handler called> () at /lib/x86_64-linux-gnu/libpthread.so.0
    #6  0x000055cf5ff65895 in hash_table_index (tbl=0x7f0efd47d9d0, tbl=0x7f0efd47d9d0, key=208) at id_table.c:136
    #7  rb_id_table_lookup (tbl=tbl@entry=0x7f0efd47d9d0, id=id@entry=3329, valp=valp@entry=0x7ffe292bef28) at id_table.c:230
    #8  0x000055cf5ffb154f in vm_search_cc (ci=0xd0100100001, klass=139702356286280) at vm_insnhelper.c:1980
    #9  rb_vm_search_method_slowpath (ci=0xd0100100001, klass=139702356286280) at vm_insnhelper.c:2077
    #10 0x000055cf5ffb1917 in vm_search_method_slowpath0 (cd_owner=139702216605640, klass=<optimized out>, cd=<optimized out>, cd=<optimized out>) at vm_insnhelper.c:2099
    #11 0x000055cf5ffccc9e in vm_search_method_fastpath (klass=<optimized out>, cd=<optimized out>, cd_owner=<optimized out>) at vm_insnhelper.c:2165
[...]

PR #191, opened 2023-03-30
Copy filter arguments to stack

peterzhu2118:
There's a bug in 50f65e42e960d7bb47d6ba4711d88bdf9280425a. The stack of the liquid VM does get modified by the filter, so it's not safe to call vm_stack_pop_n after invoking the filter.


dylanahsmith (approved):
> The stack of the liquid VM does get modified by the filter, so it's not safe to call vm_stack_pop_n after invoking the filter.

I'm sorry for leading you astray in https://github.com/Shopify/liquid-c/pull/190#discussion_r1152343829 in thinking that was safe, but I still don't understand why it isn't safe.  I wasn't concerned about the filter modifying the stack by appending to it, since I assumed it would need to pop those values back off the stack before returning.  If the stack gets reallocated, then that should update the data pointed to by `vm`, which is already needed for the `vm_stack_push(vm, result);` line.

However, I'm fine with doing something safer to try to get the crashes fixed, which should at least help validate your theory of what the problem is with the code.

casperisfine:
@dylanahsmith it's unclear to us as well as to how the stack get modified exactly, but we were able to reproduce it in the `liquid-c` test suite.

Perhaps, you'd be able to figure out what modify it? @peterzhu2118 can you share the branch with the modification check?

PR #193, opened 2023-03-31
Enable assertions when DEBUG=true

peterzhu2118:
[ruby/assert.h](https://github.com/ruby/ruby/blob/9e678cdbd054f78576a8f21b3f97cccc395ade22/include/ruby/assert.h#L36-L41) disables assertions if Ruby was not compiled with `-DRUBY_DEBUG`. We can force assertions to be ran by explicitly setting `-DRUBY_DEBUG`.

### DIFF
--- a/ext/liquid_c/extconf.rb
+++ b/ext/liquid_c/extconf.rb
@@ -12,6 +12,10 @@ end
 if ENV["DEBUG"] == "true"
   append_cflags("-fbounds-check")
   CONFIG["optflags"] = " -O0"
+  # Hack to enable assertions since ruby/assert.h disables assertions unless
+  # Ruby was compiled with -DRUBY_DEBUG.
+  # https://github.com/ruby/ruby/blob/9e678cdbd054f78576a8f21b3f97cccc395ade22/include/ruby/assert.h#L36-L41
+  $CFLAGS << " -DRUBY_DEBUG"
 else
   $CFLAGS << " -DNDEBUG"
 end


ext/liquid_c/extconf.rb
```diff
@@ -12,6 +12,10 @@
 if ENV["DEBUG"] == "true"
   append_cflags("-fbounds-check")
   CONFIG["optflags"] = " -O0"
+  # Hack to enable assertions since ruby/assert.h disables assertions unless
+  # Ruby was compiled with -DRUBY_DEBUG.
+  # https://github.com/ruby/ruby/blob/9e678cdbd054f78576a8f21b3f97cccc395ade22/include/ruby/assert.h#L36-L41
+  $CFLAGS << " -DRUBY_DEBUG"
```
dylanahsmith, line 18: Do we actually want to enable all the ruby assertions?  I think those are more for debugging ruby issues, rather than debugging ruby extension issues.  I've opened https://github.com/Shopify/liquid-c/pull/194 as a possible alternative

dylanahsmith: Oops, maybe I was too quick to merge #194.  Looks like `bundle exec rake` time difference in CI is 18s for that PR vs 20s here, so maybe the extra checks are worth it.  It wouldn't mean that ruby itself is compiled with RUBY_DEBUG, which is probably why the difference isn't that significant.

PR #195, opened 2023-03-31
Rename vm.c to liquid_vm.c

peterzhu2118:
It's confusing to debug core dumps because Ruby also has vm.c so it becomes hard to tell which frames come from Ruby and which ones come from liquid-c.


dylanahsmith:
Doesn't it end up showing the path to the shared library as well as the file in the C level backtrace information? `/tmp/bundle/ruby/3.2.0/bundler/gems/liquid-c-e1cad818edff/lib/liquid_c.so(vm_render_until_error+0x2d) [0x7fdb9e20ddd2] vm.c:441`.  Or are you seeing it elsewhere without that context?

dylanahsmith:
My concern being that the same argument could be made about many of the other files (e.g. variable.c), which could lead to wanting to prefix all the C source files.

dylanahsmith (approved):

peterzhu2118:
> My concern being that the same argument could be made about many of the other files (e.g. variable.c), which could lead to wanting to prefix all the C source files.

Fair point, another alternative would be to prefix all liquid functions with `liquid_`.

dylanahsmith:
Renaming files seems preferable than renaming functions.  Luckily, it looks like there aren't that many conflicts in practice with ruby itself and maybe the vm.c file is especially common to end up in backtraces.

I wonder how hard it would be to get our crash reporter to provide that extra information when sending it to bugsnag.

PR #199, opened 2023-07-10
only allow ruby regex whitespace chars in raw tag delimiters

ggmichaelgo:
### What are you trying to solve?
Liquid-C does not parse `raw` tag's delimiter like Liquid.
Liquid does not allow non-word characters to be part of the delimiter, but Liquid-C does.

For an example, Liquid-C should raise a syntax error for this Liquid template:
```liquid
{% raw %}

{% 🔥endraw %}
```
Since `🔥endraw` is an invalid delimiter for the `raw` tag, Liquid-C should raise a syntax error like Liquid:
```
Liquid syntax error: 'raw' tag was never closed (Liquid::SyntaxError)
```

### DIFF
--- a/ext/liquid_c/raw.c
+++ b/ext/liquid_c/raw.c
@@ -31,9 +31,7 @@ static bool match_full_token_possibly_invalid(token_t *token, struct full_token_
         char c = str[i];
 
         // match \s
-        bool is_whitespace = (
-            c == ' ' || c == '\t' || c == '\r' || c == '\n' || c == '\v' || c == '\f'
-        );
+        bool is_whitespace = rb_isspace(c);
 
         if (!is_word_char(c) && !is_whitespace && c != '%' && c != '-') {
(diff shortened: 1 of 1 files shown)


ext/liquid_c/raw.c
```diff
@@ -30,10 +30,18 @@ static bool match_full_token_possibly_invalid(token_t *token, struct full_token_
     for (long i = len - 3; i >= 0; i--) {
         char c = str[i];
 
-        if (is_word_char(c)) {
+        // match \s
+        bool is_whitespace = (
+            c == ' ' || c == '\t' || c == '\r' || c == '\n' || c == '\v' || c == '\f'
+        );
```
peterzhu2118, line 36: We can use the `ISSPACE` macro (which wraps the C function [`isspace`](https://devdocs.io/c/string/byte/isspace)) instead.

PR #211, opened 2024-01-22
Fixes for TruffleRuby and add TruffleRuby in CI

eregon:
See the individual commits for details.

Needs https://github.com/oracle/truffleruby/pull/3400 for the truffleruby-head job to pass the test suite.
(it can already be merged though, as you prefer)

PR #213, opened 2024-01-23
Release 4.2.0

casperisfine:
Ref: https://github.com/Shopify/liquid-c/pull/211

I don't have a lot of context with `liquic-c`, but based on the README it seems users are epxected to track the git gem anyway?

This release would be useful for the `yjit-bench` project.

@ggmichaelgo you seem to be the owner, is this OK with you?


ggmichaelgo (approved):
Thank you @casperisfine !
I think we should start releasing the versions with CI 🤔

PR #217, opened 2024-03-11
Add a timeout in CI and enable verbose tests for truffleruby

eregon:
* To investigate hangs like https://github.com/Shopify/liquid-c/actions/runs/8191329021/job/22400306482

### DIFF
--- a/ext/liquid_c/liquid_vm.c
+++ b/ext/liquid_c/liquid_vm.c
@@ -81,9 +81,10 @@ static void write_fixnum(VALUE output, VALUE fixnum)
     long capacity = rb_str_capacity(output);
 
     if (new_size > capacity) {
-        do {
-            capacity *= 2;
-        } while (new_size > capacity);
+        capacity *= 2;
+        if (new_size > capacity) {
+            capacity = new_size;
+        }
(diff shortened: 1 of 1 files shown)


eregon:
Running locally on truffleruby-dev, the test that hangs seems to be various tests of `VariableTest` like:
```
VariableTest#test_write_fixnum =
VariableTest#test_filter_with_variable_arg_after_const_arg =
VariableTest#test_filter_with_const_arg =
```

peterzhu2118:
> The question is whether rb_str_capacity(str) can return 0 or not.

I agree, this is a bug in liquid-c. This should be pretty simple to fix, can you send a PR?

eregon:
Confirmed that
```ruby
  describe "rb_str_capacity" do
    it "returns 0 for empty String" do
      @s.rb_str_capacity("").should == 0
    end

    it "returns 1 for 'a'" do
      @s.rb_str_capacity("a").should == 1
    end
  end
```
passes on truffleruby but fails on CRuby which returns `15` for both cases (due to embedded String).

peterzhu2118 (approved):

PR #226, opened 2024-08-20
Fix type of struct value

piperswe:
The current type for const_ptr is causing builds with Debian's GCC to
fail on 32-bit architectures, because it is being used as `VALUE *` but
is defined as `size_t *`.


piperswe:
never mind, this might be more complicated than just changing the type. I'll just disable the warning locally.

